Issue #34, opened 2015-07-30
Improve contribution guidelines to the repo


vpavlin:
As we talked on Environment & Stacks WG meeting

http://meetbot.fedoraproject.org/fedora-meeting-2/2015-07-30/env-and-stacks.2015-07-30-12.00.html/

There should probably be some "rules" enforced during content addition.
1. Every section should have editor - i.e. subject matter expert has to review all content changes for the given section
2. Before starting work on a specific topic, issue has to be created and draft should be linked from this issue so that editors and others can review and comment the draft early
3. Every content change needs to be proposed as a Pull Request and needs to get LGTM from two editors (if the PR submitter is also an editor, only one LGTM from another editor is ok to merge)

I guess, at least in the beginning, it will be hard to get editors for all the content, but it should definitely be one of the goals.

asamalik:
+1

strzibny:
I mostly agree.

strzibny:
We started to work on the guidelines: https://github.com/developer-portal/content/blob/master/CONTRIBUTING.md.

jackorp:
Current Contribution guidelines reside here: https://developer-portal.github.io/.

Feel free to reopen this issue if further work is needed.

PR #35, opened 2015-08-03
Added Docker installation tutorial prototype

vjancik:
**Problems:**

- Multiline code segments aren't getting split properly (on line end), thus all lines look like one line on the website.
- Adding `.md` pages to `content/tools/docker` results in new entries on the *Tools* page on the website and not the intended additional page under *Docker* article. Maybe I am doing this wrong. The consequence of this is, I can't split my Docker documentation to multiple pages.

**Notes:**

- The installation tutorial is basically an up to date rewrite of [Official Docker Fedora Installation Guide](https://docs.docker.com/installation/fedora/)

**Ideas:**

- For the *Create your first image* section, I'd like to rework the [Official Docker Create your first image tutorial](http://docs.docker.com/linux/step_four/) to use a Fedora based image (which I will create) with the same outcome. What do you think?

**Need feedback:**

- I wasn't sure what to write in the *What is Docker* section, that should precede the *Installation* section, because I don't know how technical I should make it. E.g. whether I should write about cgroups, namespaces and chroot.
- I currently don't have enough experience with Docker to write a comprehensive *Creating a Dockerfile* section. I will learn it in the course of the next few weeks, but I wanted to let you know, in case you'd need to move faster, thus assign it to someone else.


Vixian:
**Edit:** 
- It seems *docker* package is only available for Fedora 22 and higher right now. Should I add the upstream script `curl -sSL https://get.docker.com/ | sh` which adds upstream docker repository to yum as the right way to install it on Fedora 20&21?

phracek:
I don't thing that upstream script is the good approach. It would be good also to add at the end usage from DevAssistant point of view like da tweak docker develop etc.

This can be advise by bkabrda.

I will ask coleagues who can help us with review.

tools/docker/installation.md
```diff
@@ -0,0 +1,89 @@
...
+###**Installing Docker on Fedora**
+
+####**Prerequisites**
+
+Docker is currently supported on the following versions of Fedora:
+
+* Fedora 20
+* Fedora 21
+* Fedora 22
+
```
bkabrda: So why don't we just say that it's been supported since Fedora 20? That way, it would be unnecessary to update the page on every Fedora release.

tools/docker/installation.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+You have to log out and log back in for these changes to take effect. Then you can verify if your changes were successful by running Docker without `sudo`.
+
+####**Start the Docker daemon at boot**
+
+To make Docker start when you boot your system, use the command:
+
+```
+$ sudo systemctl enable docker
```
bkabrda: I'd recommend to move this to `sudo systemctl start docker` and keep these two together. It makes more sense to me that way.

tools/docker/installation.md
```diff
@@ -0,0 +1,89 @@
...
+
+```
+$ sudo docker run hello-world
+```
+
+####**Creating a Docker group**
+
+The Docker daemon binds to a Unix socket instead of a TCP port. By default that Unix socket is owned by the user `root` and other users can access it with `sudo`. For this reason, Docker daemon always runs as the `root` user.
+
+To avoid having to use `sudo` when you use the Docker command, create a Unix group called Docker and add users to it. When the Docker daemon starts, it makes the ownership of the Unix socket read/writable by the Docker group.
```
bkabrda: I think you should call the group `docker`, not Docker (to suggest that it's a name that one would use when creating the group on command line); there are multiple occurences of this, so please fix them all if you decide to follow this advice.

tools/docker/installation.md
```diff
@@ -0,0 +1,89 @@
...
+
+```
+$ uname -r
+4.0.8-300.fc22.x86_64
+```
+
+If your kernel version is less than 3.10, you can update it using the following command:
+
+```
+$ sudo dnf update kernel
```
TomasTomecek: Isn't this too thorough?

tools/docker/installation.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+You need to restart your computer to boot with the new kernel version.
+
+####**Installation**
+
+The easiest way to install Docker on Fedora is from a Fedora package:
+
+```
+$ sudo dnf install docker
```
TomasTomecek: Since you mentioned F2{0,1,2}, it would be worth saying, that on F2{0,1} the package is named docker-io and [docker](http://koji.fedoraproject.org/koji/buildinfo?buildID=442722) is something completely different.

tools/docker/installation.md
bkabrda: Good point. I think that currently only F21 and later are supported releases and these have (AFAIK) always shipped with kernel >= 3.10.

tools/docker/installation.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+To start the Docker service use:
+
+```
+$ sudo systemctl start docker
```
TomasTomecek: ...also, why `sudo`? is this best practice for whole portal?

tools/docker/installation.md
```diff
@@ -0,0 +1,89 @@
...
+$ sudo systemctl enable docker
+```
+
+For additional Systemd configuration options for Docker, like adding an HTTP Proxy, reffer to Docker documentation [Systemd article](https://docs.docker.com/articles/systemd/)
+
+###**Creating your first image**
+
+TODO
+
+###**Writing a Docker file**
```
TomasTomecek: *Dockerfile

tools/docker/installation.md
```diff
@@ -0,0 +1,89 @@
...
+
+TODO
```
TomasTomecek: I would also add troubleshooting guide.

tools/docker/installation.md
strzibny: I agree, it should be `docker`.

TomasTomecek:
> I wasn't sure what to write in the What is Docker section, that should precede the Installation section,
> because I don't know how technical I should make it. E.g. whether I should write about cgroups,
> namespaces and chroot.

I advise not to go into technical details. Plain and simple introduction should be sufficient -- *why would you use docker*?

> `curl -sSL https://get.docker.com/ | sh`

please don't even mention this; `curl ... | sh` is the worst practice ever; every time someone writes this, a baby sheep dies

strzibny:
@phracek I am also thinking how to best integrate notes on DevAssistant assistants as this would be a case for Vagrant content as well. Perhaps this should be made in a similar fashion across the whole content. The question is how.

tools/docker/installation.md
phracek: you mean like switching to root and call all commands directly?

tools/docker/installation.md
phracek: Good point:)

phracek:
@strzibny Yeah. We have decided (together with Radek Vokal) that each section which already has something with DevAssistant should be mentioned here.

phracek:
@Vixian Could you please a rewrite it a bit on the base of our comments?

phracek:
@Vixian: Would you be able to create a docker example how to create them and execute them on the system? It should be a really simple example.

tools/docker/installation.md
strzibny: We should agree on one way or another. I used `#` for root commands and `$` for user commands in my content.

TomasTomecek:
> @Vixian: Would you be able to create a docker example how to create them and execute them on the
> system? It should be a really simple example.

To elaborate a bit, I think the guide should show why users should use docker — how it can ease the pain of setting up, maintaining and sharing a development environment.

The example could be that you are coding your app on your host and you are bindmounting app's sources inside container where you have webserver running in your production environment.

(I would actually like to see a `docker-compose` example, but since we don't have it in fedora still... I could write such sample myself)

strzibny:
> To elaborate a bit, I think the guide should show why users should use docker — how it can ease the pain of setting up, maintaining and sharing a development environment.

> For the Create your first image section, I'd like to rework the Official Docker Create your first image tutorial to use a Fedora based image (which I will create) with the same outcome. What do you think?

In my opinion let's avoid this. Our job is not to explain Docker the same way as I am not gonna explain Vagrant or what Ruby is. This guides are about what we have in Fedora and how to get it running (install package, set `env` variable) and nothing else. We are not here to duplicate upstream guides, nor to sell it to our users. The usage should be IMHO explained only when Fedora differs from upstream (e.g. as in Ruby content that you need to explicitly require JSON). Otherwise let upstream documentation work for us. Let's just document Fedora specifics.

Think also about future maintenance cost if you add things like this one. Who is gonna make sure that the content on creating your Dockerfile is current? Let's just include links to upstream guides ESPECIALLY if they do exist.

> Adding .md pages to content/tools/docker results in new entries on the Tools page on the website and not the intended additional page under Docker article. Maybe I am doing this wrong. The consequence of this is, I can't split my Docker documentation to multiple pages.

This does not matter, we fix that later, split the content freely.

vpavlin:
I'd suggest to add a section __Featured images__ or something like this and mention 

* Fedora base image, 
* where to get Rawhide image (currently the simplest way to get it is use this link: http://koji.fedoraproject.org/koji/tasks?state=all&view=flat&method=createImage&order=-id and find first successfully built Docker image..)
* maybe link to [Fedora-Dockerfiles](https://github.com/goern/Fedora-Dockerfiles/) repository

Also definitely a link to Vagrant section with description stating something like: 
> If you don't want to install docker on your machine, you can try it in VM easily with [Vagrant](http://link.to/vagrant/section) and [Project Atomic box](http://link.to/project/atomic/box)

vjancik:
I apologize for the delay, I was busy with Red Hat work and on a vacation for a week.

I implemented your suggestions. Now could you clarify, whether we are keeping `$ sudo` or using `#`?

Also, for the basic usage tutorial, I am hearing mixed opinions. Should I write a basic tutorial, or should I link to the upstream tutorial from the docs?

And lastly, what is our timeline? I prefer to work with a deadline.

strzibny:
> I implemented your suggestions. Now could you clarify, whether we are keeping $ sudo or using #?

Use `#` now. If we decide later otherwise we go through everything and change it.

> Also, for the basic usage tutorial, I am hearing mixed opinions. Should I write a basic tutorial, or should I link to the upstream tutorial from the docs?

Please link the docs.

> And lastly, what is our timeline? I prefer to work with a deadline.

End of summer?

TomasTomecek:
@strzibny 

I disagree. 

If I look at [project desciption](https://fedoraproject.org/wiki/Websites/Developer) of what Developer Portal is, I can see a lot of sentences like *"How to use this tech?"*, *"...a new place for developers, providing information about tools, technology..."*, *"Discovering what else Fedora offers and learning how to use it."*

In the end I agree that we should **NOT** provide advanced tutorials, but why not some basic, end-to-end guide which captures some developer's workflow executed on top of Fedora stack (not many upstreams use Fedora as an example: [search for ubuntu and fedora](https://docs.docker.com/reference/builder/)).

Maybe it's time to write [a guide for contributors](https://github.com/developer-portal/content/blob/master/CONTRIBUTING.md).

As I'm not invloved in this project, don't take my ideas seriously.

Good points, @vpavlin. Those should be definitely mentioned IMO.

vpavlin:
@TomasTomecek Yes, contribution guidelines already have their own issue - #34 :)

strzibny:
Please mention Fedora Dockerfiles as per https://github.com/developer-portal/content/pull/41

phracek:
Hi folks, what is blocking this issue for merge?

vpavlin:
@phracek Some changes were requested but not delivered - Fedora Dockerfiles mainly, IMO

strzibny:
(And there are conflicts so we can't merge.)

phracek:
@vjancik  Hi Victor, could you please update docker/installation.md with comments mentioned above?
I did not see it in last pull request?

PR #36, opened 2015-08-04
Initial PHP Content

frantisekz:
I will appreciate your feedback and I hope I didn't make any serious mistakes :)

It's probably short, but enough to get PHP working and start developing with it.

### DIFF
--- a/tech/languages/php/php-installation.md
+++ b/tech/languages/php/php-installation.md
@@ -1,0 +1,65 @@
+---
+title: PHP
+page: php
+section: tech-languages
+---
+
+## PHP Installation
+
+### PHP
+
+To install PHP, simply type:
+
+```
+# dnf install php
+```
+
+Above command will install latest stable PHP
+packages including some additional modules. 
+
+### Web server
+
+To fully make use of PHP you will probably
+want to install web server. 
+Apache is one of the recommended and easy to
+set up. To install Apache, type:
+
+```
+# dnf install httpd
+```
+
+Above command will install latest Apache web server.
+Default working directory is /var/www/html.
+
+To enable Apache autostart on boot, type:
+
+```
+# systemctl enable httpd.service
+```
+
+To start Apache, type:
+
+```
+# systemctl start httpd.service
+```
+
+### MySQL server
+
+PHP applications can use MySQL server to store data.
+To install MySQL server (Mariadb) type:
+
+```
+# dnf install mariadb mariadb-server php-mysql
+```
+
+To enable Mariadb autostart on boot:
+
+```
+# systemctl enable mariadb.service
+```
+
+To start Mariadb, type:
+
+```
+# systemctl start mariadb.service
+```
--- a/tech/languages/php/php-installation.md
+++ b/tech/languages/php/php-installation.md
@@ -11,55 +11,54 @@ section: tech-languages
 To install PHP, simply type:
 
 ```
-# dnf install php
+# dnf install php-cli
 ```
 
-Above command will install latest stable PHP
-packages including some additional modules. 
+Above command will install latest stable PHP package, which includes the command-line interface executing PHP scripts, /usr/bin/php, and the CGI interface. 
 
-### Web server
 
-To fully make use of PHP you will probably
-want to install web server. 
-Apache is one of the recommended and easy to
-set up. To install Apache, type:
+You might want to install some other packages, e.g. phpunit for unit tests or composer to manage dependencies of PHP projects. You can install them with:
 
 ```
-# dnf install httpd
+# dnf install phpunit composer
 ```
 
-Above command will install latest Apache web server.
-Default working directory is /var/www/html.
-
-To enable Apache autostart on boot, type:
+### PHP modules
+PHP provides only a few set of extensions, other should be installed separately. For example, if you need "mysqli" extension, use:
 
 ```
-# systemctl enable httpd.service
+# dnf install php-mysqli
 ```
 
-To start Apache, type:
+### PEAR
+
+The PHP Extension and Application Repository is a repository of PHP software code. 
+To install PEAR, type:
 
 ```
-# systemctl start httpd.service
+# dnf install php-pear
 ```
-
-### MySQL server
-
-PHP applications can use MySQL server to store data.
-To install MySQL server (Mariadb) type:
+Use pear command to install PEAR extensions. To install Mail_Mime package (for example), type:
 
 ```
-# dnf install mariadb mariadb-server php-mysql
+# pear install Mail_Mime
 ```
 
-To enable Mariadb autostart on boot:
+You can [find more PEAR modules](http://pear.php.net/packages.php).
+
+
+### Frameworks
+
+We have packaged two popular PHP frameworks in Fedora: Zend and Symfony. 
+
+To install Zend framework, type:
 
 ```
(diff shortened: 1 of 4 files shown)


strzibny:
Hi,

the PHP content should probably stick to PHP. We will cover databases separately. I would also avoid the `httpd` part, we could create something like stacks later on (and on other place) where we could mention LAMP stack as a complete solution.

In my opinion this guide should focus on installing PHP (that's done), installation of PEAR and PEAR extensions (you can take a look how I covered RubyGems for Ruby) and most popular frameworks we have packaged in Fedora (probably Zend, Symfony).

If you can rework this in this way that would be absolutely great.

asamalik:
I agree. There is a section about creating web applications and LAMP stack could be one of the options. 

If you want to create a content just about PHP, continue with this pull-request.
If you want to write about creating a web application using the LAMP stack, you can create a page in the Web Application section which is currently under /start/sw/web-app directory in this repo.

And thank you for contribution!

phracek:
@frantisekz Could you please add a section about PEAR. Ruby content has also an information about RoR. Thanks.
php-installation.md file and send us an update. Thanks, though.

phracek:
@remicollet Could you please review it a bit more? Thanks
We would like to inform user how to use PHP in Fedora for Developers.

remicollet:
Quick notes:

    # dnf install php 

"php" Have no sense ;) This will install various stuff which is probably not what user want (especially mod_php and httpd), and this is not the goal (see Web server later)

    To install PHP command, simply type: 
    # dnf install php-cli


And later in the Web Server section

     # dnf install httpd mod_php
     or
     # dnf install httpd php-fpm
     or
     # dnf install nginx php-fpm

The 3 works now out of the box (of course the relevant services need to be started)

As @phracek  said a note about PEAR could have sense, but in this case I will prefer to enlight some developper tools (instead of PEAR which is quite old / deprecated)

     Developper could find some value in installing some additional tools, e.g.
     # dnf install php-pear  composer   phpunit   phpcompatinfo

And some note about  extensions installation

     PHP provides only a few set of extensions, other should be installed separately, usually, if you need "foo" extension
    # dnf install php-foo
    eg
    # dnf install php-mbstring  php-mysql  php-xmlwriter  php-mongo

remicollet:
And also a note about IDE could be useful (e.g. eclipse-pdt)

strzibny:
@remicollet Great feedback, that's exactly why I think that this portal makes sense and the difference between installing php and php-cli should be explained. Regarding the IDEs we so far dismissed the idea, although we can revisit that in future in we have everything else.

frantisekz:
Thanks for the feedback. I will update this PR tomorrow.

frantisekz:
I've updated this according to feedback.

@remicollet I've omitted example "foo" extension - I believe it's better not to include commands that will produce errors. What about others on this?

Should I keep everything in one file, or split it up?

remicollet:
Too much light on "pear". Especially, explaining how to install package using it, when so much are available as RPM (see http://rpms.remirepo.net/rpmphp/rpm.php?type=pear)

php-ZendFrammework is the old deprecated v1, php-ZendFrammework2 should be prefered. But having information about such Framework, without any explanation how to use it...

To be clear, today PHP dev only use composer.

remicollet:
Perhaps this page should at least answer the basic question all PHP dev ask
* where do I put my app so it will be managed in the webserver

Perhaps, the "simple" solution using the integrated PHP Developement Server worth to be added

In the project tree

     php --server localhost:8080 --docroot  .

And then browse  http://localhost:8080/

frantisekz:
@remicollet Thanks for you feedback again.

I am going to start guide for web applications. That was intended at first, but I've misunderstood that the language section wasn't for Web Apps. Anyway, I'll try to fix and finish this PR too.

frantisekz:
I believe I've improved this according to your feedback. The only area I doubt I can describe properly are Frameworks - I've never used any of them. So IMO it'd be better to pull this only with frameworks installation and somebody else can finish that part. 
Since @remicollet is away, can you ( @phracek @asamalik @strzibny  ) take a look please? 

Many thanks

tech/languages/php/php-installation.md
```diff
@@ -0,0 +1,59 @@
...
+
+### PHP Installation
+
+To install PHP, simply type:
+
+```
+# dnf install php-cli
+```
+
+Above command will install latest stable PHP package, which includes the command-line interface executing PHP scripts, /usr/bin/php, and the CGI interface.
```
strzibny: I would say "`/usr/bin/php` executable" and also use backticks.

tech/languages/php/php-installation.md
```diff
@@ -0,0 +1,59 @@
...
+### PHP Installation
+
+To install PHP, simply type:
+
+```
+# dnf install php-cli
+```
+
+Above command will install latest stable PHP package, which includes the command-line interface executing PHP scripts, /usr/bin/php, and the CGI interface. 
+
```
strzibny: Redundant line.

tech/languages/php/php-installation.md
```diff
@@ -0,0 +1,59 @@
...
+
+You might want to install some other packages, e.g. phpunit for unit tests or composer to manage dependencies of PHP projects. You can install them with:
```
strzibny: Use backticks around names of binaries such as phpunit and composer. Or use their proper name: Composer (capital C).

tech/languages/php/php-installation.md
```diff
@@ -0,0 +1,59 @@
...
+
+```
+# dnf install phpunit composer
+```
+
+### PHP modules
```
strzibny: Put extra line after headline.

tech/languages/php/php-installation.md
```diff
@@ -0,0 +1,59 @@
...
+PHP provides only a few set of extensions, other should be installed separately. For example, if you need "mysqli" extension, use:
+
+```
+# dnf install php-mysqli
```
strzibny: Is every PHP extension packaged named php-NAME? If so, please write it the way this is clear. Is there also some virtual provide users can use to find/install PHP extensions?

tech/languages/php/php-installation.md
```diff
@@ -0,0 +1,59 @@
...
+
+You can start PHP Development Server by typing this in your project's root:
+
+```
+# php --server localhost:8080 --docroot  .
+```
+Server will be accessible on http://localhost:8080/ .
+
+
+### Frameworks
```
strzibny: I would make a separate page for each framework as they will help people to navigate on the site. Nobody will look on how to install Zend on the page with "PHP installation" title. Is it also common in the community to install these frameworks using composer? Perhaps both ways could be showed (as I did with installing Ruby on Rails).

strzibny:
Thanks @remicollet and @frantisekz for working on this! I included some inline comments.

frantisekz:
Zend Framework 2 with Composer is possible but it isn't as easy as RoR is, take a look: https://packages.zendframework.com/

Shall I include it?

Anything else should be updated according to your notes.

strzibny:
In that case I think it's not necessary, at least for now.

asamalik:
Thanks a lot for contribution!

phracek:
@frantisekz : Please send me to my email address phracek@redhat.com your contact information. I will send info about Cool Stuff. Thanks.

PR #38, opened 2015-08-08
Description of C and C++ installation

mruprich:
I tried to write a little something about running C and C++ under Fedora. Please tell me if you like it or tell me what should I change or add or remove.


strzibny:
My feedback:

* I would split C and C++ content
* I would mention both `gcc` and `clang` installation

phracek:
I would be also glad if you are able to describe a bit how automake and cmake works.
@kdudka Could you please review it?

tech/languages/c/about.md
```diff
@@ -4,5 +4,67 @@ page: c
...
+### C
+
+C is a compiled language and as such it needs a compiler. To install a compiler, simply type:
+
+```
+# sudo dnf install gcc
+```
+
+This command should download and install package with compiler called gcc.
+To compile a program written in C, simply run the gcc like this:
```
kdudka: To compile and link... (compilation only mode can be enabled by the -c option of gcc)

tech/languages/c/about.md
```diff
@@ -4,5 +4,67 @@ page: c
...
+# man gcc
+```
+
+### C++
+
+Like C, C++ is also a compiled language. It has many similarities with C and to some extent it is safe to say that C is
+subset of C++ and C++ supports every programming technique supported by C. Next command shows how to install C++ compiler:
+
+```
+# sudo dnf install g++
```
kdudka: ```# sudo dnf install gcc-c++```

tech/languages/c/about.md
```diff
@@ -4,5 +4,67 @@ page: c
...
+```
+
+To compile your program you can do the same as was described above with C:
```
kdudka: To compile and link ....

strzibny:
I would probably prefer to write \`file\` at the beginning of sentence rather than File (with capital F changing the name of the file or abbreviation). However I am unsure what's the right way to do it in English.

PR #41, opened 2015-08-26
contributing: tie Fedora-Dockerfiles to content

TomasTomecek:


strzibny:
Definitely a good idea. We should probably mention it the docker section. 

I am trying to think about a good way ho to mention it in the sections themselves. Do you know if these images are available in Docker Hub?

TomasTomecek:
I can see [there is a Fedora namespace](https://hub.docker.com/u/fedora/) on docker hub. After a brief check, it looks like that some images are pretty outdated. Not sure who maintains those.

asamalik:
We have moved our contributor guidelines to developer-portal.github.io

PR #42, opened 2015-08-27
Initial content for Java

msrb:
Initial "something" for Java. It's marked as work-in-progress for now, because I would like to have it reviewed by other Java SIG members first.

### DIFF
--- a/tech/languages/java/java-build-tools-installation.md
+++ b/tech/languages/java/java-build-tools-installation.md
@@ -1,0 +1,32 @@
+---
+title: Build Tools
+page: java
+order: 2
+---
+The most popular build tools used by millions of Java developers are packaged in official Fedora repositories. This page contains information on how easy it is to install them.
+
+## Maven installation
+
+Maven is probably the most popular build tool in Java world these days. To install it, simply type:
+
+```
+$ sudo dnf install maven
+```
+
+## Gradle installation
+
+Gradle is another popular tool used not only for building Java projects. To install it, simply type:
+
+```
+$ sudo dnf install gradle
+```
+
+## Ant+Ivy installation
+
+Last but not least, still quite popular duo Ant+Ivy can be installed simply by typing:
+
+```
+$ sudo dnf install ant apache-ivy
+```
+
+<!-- TODO: once content for Scala programming language is created, mention here that it's also possible to use SBT to build Java projects + link to corresponding Scala section. -->
--- a/tech/languages/java/java-sig.md
+++ b/tech/languages/java/java-sig.md
@@ -1,0 +1,11 @@
+---
+title: Java SIG
+page: java
+order: 3
+---
+
+## Java SIG
+
+[Java SIG](https://fedoraproject.org/wiki/SIGs/Java) is a special interest group within Fedora Project. The group focuses on packaging Java libraries and applications, setting and improving standards for packaging and collaboratively managing bigger changes related to Java in Fedora.
+
+If you are interested in Java and Fedora, join us!
--- a/tech/languages/java/about.md
+++ b/tech/languages/java/about.md
@@ -1,9 +1,0 @@
----
-title: Java
-page: java
-section: tech-languages
-version: 1.8.0
----
-
-# Java
-This content needs to be created...
--- a/tech/languages/java/java-installation.md
+++ b/tech/languages/java/java-installation.md
@@ -1,0 +1,40 @@
+---
+title: Java
+page: java
+order: 1
+section: tech-languages
+---
+
+## Java installation
+
+Fedora comes with OpenJDK - a free and open source implementation of the Java Platform, Standard Edition. To install it, simply type:
+
+```
+$ sudo dnf install java-1.8.0-openjdk-devel
+```
+
+This command will install Java runtime environment and associated development tools.
+
+IcedTea-Web may come handy if your plan is to develop Java applets or to experiment with Java Web Start. To install it, simply type:
+
+```
+$ sudo dnf install icedtea-web
+```
+
+This command will install Java web browser plugin and free implementation of Java Web Start.
+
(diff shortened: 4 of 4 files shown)


tech/languages/java/java-sig.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Java SIG
+page: java
+---
+
+## Java Special Interest Group
+
+[Java SIG](https://fedoraproject.org/wiki/SIGs/Java) is a special interest group within Fedora Project. The group focuses on packaging Java libraries and applications, setting and improving standards for packaging and collaboratively managing bigger changes related to Java in Fedora.
+If you are interested in Java and Fedora, join us!
```
strzibny: Perhaps a new line above?

tech/languages/java/java-build-tools-installation.md
```diff
@@ -0,0 +1,28 @@
+---
+title: Java Build Tools
+page: java
+---
```
strzibny: We now support `order` paramer, so please add it and order the pages as:
Java installation: order: 1
Build tools: order: 2
Java SIG: order: 3

tech/languages/java/java-build-tools-installation.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Maven installation
+
+Maven is probably the most popular build tool in Java world these days. It's available in official Fedora repositories, in form of RPM. To install it, simply type:
+
+```
+# dnf install maven
+```
```
strzibny: I miss an example of installing a package from this build tools, also, are some of the libraries from Maven repositories packaged in Fedora?

strzibny:
I miss a page about installing `springframework` package. And also, are there other popular frameworks that people use? Perhaps not only for web, but mobile and desktop.

tech/languages/java/java-build-tools-installation.md
msrb: We have libraries from Maven Central Repository in Fedora, but they are not used by developers (or they are used, but very very rarely). Trying to explain here how the packaged artifacts should be used during development could only confuse developers (the process is not straightforward).

msrb:
Adding section about mobile development is a good idea. But for ```springframework``` applies what I said above about packaged libraries from Maven repositories.

tech/languages/java/java-build-tools-installation.md
msrb: will do

tech/languages/java/java-sig.md
msrb: Not sure if I am following here. What exactly do you have in mind? :)

strzibny: paragraph

paragraph

e.g. "if you are inte..." looks like a new paragraph, no?

tech/languages/java/java-build-tools-installation.md
strzibny: Especially if that's not straightforward how do I learn how to do it?

strzibny:
So I can't easily use that package to start my Spring app?

tech/languages/java/java-sig.md
msrb: Good catch, I will fix it.

msrb:
Nope, there is no easy way. AFAIK it doesn't have any user-visible interface. It's just a library.

tech/languages/java/java-build-tools-installation.md
msrb: Long story short, Java is special :) Using artifacts from RPMs for development is hard and I fully understand why people don't use them. Projects like [Language Specific Repositories] could improve the situation significantly, but it's a far cry.

[Language Specific Repositories]: https://fedoraproject.org/wiki/Env_and_Stacks/Projects/LanguageSpecificRepositories

phracek:
What do you think about mobile development? Like android?

msrb:
@phracek yep, exactly. I was thinking about adding section about Android development in Fedora.

tech/languages/java/java-build-tools-installation.md
strzibny: Ok, let's forget that for now. But please mention an example of using the build tools. Just a simple one.

p.s. just wondering, is there an app in fedora that uses packaged version of Spring framework?

msrb: I am not really opposed to the idea of briefly mentioning how to use the build tools, but it's not really Fedora-specific. Maybe mentioning that after successful installation, the "mvn" command will be available on the system? I guess that the Fedora developer portal shouldn't be about teaching users how to use Maven, right?

msrb: There are applications that use packaged version of Spring framework (most prominent is probably Jenkins). That's the reason why it is in Fedora.

strzibny: > I guess that the Fedora developer portal shouldn't be about teaching users how to use Maven, right?

I am the biggest advocate of this, but perhaps the most common install command is good to mention (for instance for Python/pip and Ruby/RubyGems it makes sense especially when comparing to installation of RPMs... I know that since we won't have that for Java, it seems less important).

msrb:
Not sure why we prefer "$ sudo do something" instead of plain-old "# do something". I understand the advantages of sudo approach, but users are not in sudoers file by default in Fedora. Getting error messages like "user is not in the sudoers file. This incident will be reported." when simply copy-pasting commands from developer portal doesn't scream user-friendly, if you ask me. But so be it.

strzibny:
> Not sure why we prefer "$ sudo do something" instead of plain-old "# do something".

I think we will revisit the decision, I started with `#` anyway and I see that Fedora Documentation also uses `#` or explicitly states "run as root". `sudo` is because of the convenience of copy&paste. Either way we should probably document it.

strzibny:
Merging, feel free to submit changes or new sections later :). Thanks!

PR #48, opened 2015-09-01
Added content for Arduino, Arduino IDE and Ino tool

thozza:
Signed-off-by: Tomas Hozza <thozza@redhat.com>

### DIFF
--- a/start/hw/arduino/about.md
+++ b/start/hw/arduino/about.md
@@ -4,6 +4,52 @@ page: arduino
 section: start-hw
 ---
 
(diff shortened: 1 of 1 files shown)


thozza:
So after our discussion over IRC you seemed to be OK with the text as is. The Fedora on Arduino was more of a joke and preparation for the following part of the sentence.

thozza:
I accidentally removed the comment from the sources. However Petr had objections about the "Fedora on Arduino" part.

PR #54, opened 2015-09-02
Use only one top level heading

hhorak:
Based on comments in issue #14.


hhorak:
Actually the comments were in the pull request #53 ...

PR #59, opened 2015-09-02
add docker-compose content

TomasTomecek:
It's still not done, there are still some things I want to improve:
- [x] don't require to install django on host, use admin command from container
- [x] open PR wrt django docker image
- [x] style, typos, grammar
- [x] show how to inspect database

### DIFF
--- a/tools/docker/compose.md
+++ b/tools/docker/compose.md
@@ -1,0 +1,213 @@
+---
+title: Docker Compose
+page: docker-compose
+section: tools
+---
+
+# Docker Compose
+
+[docker-compose](https://docs.docker.com/compose/) is a tool to orchestrate docker containers. No more 4 lines long `docker run` commands, just one YAML file which describes your whole setup.
+
+In this example we'll go through a simple development environment setup:
+
+* first step will be to start a [Django](https://www.djangoproject.com/) project
+* we will bindmount sources inside a docker container (app will change as we code)
+* our choice of database will be PostgreSQL (which we link to the container with sources)
+* the whole development environment will be managed by docker-compose
+
+We will name our project `awesome_web`. So let's start by creating directory for it:
+
+```
+$ mkdir awesome_web
+$ cd awesome_web
+```
+
+Whole database will live in subdirectory `db` (postgres user has to own it -- docker still lacks user namespaces):
+
+```
+$ mkdir db
+# chown 26:26 db
+```
+
+## Prerequisites
+
+We'll use fedora-based images from [fedora-dockerfiles](https://github.com/fedora-cloud/Fedora-Dockerfiles).
+
+```
+$ git clone https://github.com/fedora-cloud/Fedora-Dockerfiles.git
+```
+
+Let's start with Django. Since we'll bindmount sources inside container, we won't use `Dockerfile` from `Fedora-Dockerfiles`, we'll improve it a bit instead:
+
+```
+FROM fedora
+MAINTAINER http://fedoraproject.org/wiki/Cloud
+
+RUN dnf -y update && dnf clean all
+RUN dnf -y install python-pip python-django git sqlite python-psycopg2 && dnf clean all
+
+# create directory /code and mount sources there
+RUN mkdir /code
+WORKDIR /code
+VOLUME /code
+
+EXPOSE 8000
+
+CMD ./manage.py runserver 0.0.0.0:8000
+```
+
+```
+$ cd Fedora-Dockerfiles/Django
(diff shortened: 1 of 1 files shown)


tools/docker/compose.md
```diff
@@ -0,0 +1,145 @@
+---
+title: Basic Usage
+page: docker-compose
+section: tools
+---
+
+We'll go through a simple [docker-compose](https://docs.docker.com/compose/) usage. What do we do?
```
strzibny: 'We will'

tools/docker/compose.md
```diff
@@ -0,0 +1,145 @@
...
+Time to spin the whole environment!
+
+```
+$ docker-compose up
+```
+
+That's it! You have your development environment running in docker containers! You can check easily:
+
+```
+elinks -dump http://$(docker inspect -f '{{ .NetworkSettings.IPAddress }}' awesomeweb_web_1):8000
```
strzibny: You still want to use `$` sign here I guess.

TomasTomecek: Yes, thanks. Will definitely read the whole thing several times before it's done and ready for merge -- this is basically still a wip.

tools/docker/compose.md
TomasTomecek: I'll reword probably. Don't want to have two "we will"s next to each other.

TomasTomecek:
done, ready for review & merge

TomasTomecek:
@strzibny thank you very much for review! Most of the comments addressed.

strzibny:
This is confusing.

> Let's start with Django. Since we'll bindmount sources inside container, we won't use Dockerfile from Fedora-Dockerfiles, we'll improve it a bit instead:

You talking about improving a Dockerfile without saying which one you mean and then you continue as:

> $ cd Fedora-Dockerfiles/Django

This should be IMHO sooner e.g. 1, let's go to Django directory 2, Let's edit Dockerfile 3, ...

@phracek Since you merged it, did you really go though? I wanted to do that today as I discussed with Tomas.

strzibny:
Also, you have `Prerequisites` section but you don't mention to install Docker (with link to Docker installation page) and you don't say that the daemon has to be running! I even think the this section should come sooner ...before even creating a project folders and such. Visitors should know what they will need before they can start with the tutorial.

TomasTomecek:
@strzibny will address those issues in new PR

strzibny:
Thanks, I will put here other things as I am going through...

> cat docker-compose.yml

You probably mean creating docker-compose.yml yet it's not clear that this needs to be created (a command with heredoc that would actually write to that file would be better). Also at this stage in the tutorial, user is at Fedora-Dockerfiles/postgresql....are you sure you want them to create the file there? Perhaps you meant awesome_web folder instead?

strzibny:
When is docker-compose to be part of fedora? The upstream installation guide is a bit confusing, perhaps we should say at the beginning: install docker (link to our guide) and then include the upstream installation here (easy steps that can be copy&pasted).

strzibny:
> docker-compose run web django-admin startproject awesome_web .

This gives me:
Starting awesomeweb_db_1...
Cannot start container 724f8cb2d227a6efa5e82d9669235430f63dadc76a5ddd4907248f1edc11490a: Cannot link to a non running container: /awesomeweb_db_1 AS /awesomeweb_web_run_2/awesomeweb_db_1


> Since we use PostgreSQL, we need change DATABASE variable in settings.py:

It's `DATABASES` and settings.py is not in my project folder (perhaps because the last command failed? I don't know, I am stuck).

TomasTomecek:
> When is docker-compose to be part of fedora? The upstream installation guide is a bit confusing, perhaps we should say at the beginning: install docker (link to our guide) and then include the upstream installation here (easy steps that can be copy&pasted).

It is already part of Fedora: https://admin.fedoraproject.org/pkgdb/package/docker-compose/

...so I omitted the most important part -- how to install it (/facepalm)

> This gives me:
```
Starting awesomeweb_db_1...
Cannot start container 724f8cb2d227a6efa5e82d9669235430f63dadc76a5ddd4907248f1edc11490a: Cannot link to a non running container: /awesomeweb_db_1 AS /awesomeweb_web_run_2/awesomeweb_db_1
```

There was something wrong with the database container. I'll add troubleshooting section.

TBH the truth is, that pgsql dockerfile is a bit outdated and we need [this](https://github.com/fedora-cloud/Fedora-Dockerfiles/pull/128) and [this](https://github.com/fedora-cloud/Fedora-Dockerfiles/pull/127) merged to make it work.

PR #64, opened 2015-09-07
improve docker-compose content

TomasTomecek:
Follow-up on #59 

@strzibny Thanks a lot for your very thorough review.

### DIFF
--- a/tools/docker/compose.md
+++ b/tools/docker/compose.md
@@ -32,7 +32,7 @@ In this example we'll go through a simple development environment setup:
 
 ### Docker
 
-This guide expects that you have docker engine installed, configured and running (if your user is not able to reach docker, prefix all `docker` commands with `sudo `; for more information see section docker of this site).
+This guide expects that you have docker engine installed, configured and running (for more information see section docker of this site).
 
 
 ### Images
@@ -65,14 +65,14 @@ CMD ./manage.py runserver 0.0.0.0:8000
 Let's build it:
 
 ```
-$ docker build --tag=fedora-django .
+# docker build --tag=fedora-django .
 ```
 
 And now PostgreSQL (stock one is just fine):
 
 ```
 $ cd Fedora-Dockerfiles/postgresql
-$ docker build --tag=fedora-postgresql .
+# docker build --tag=fedora-postgresql .
 ```
 
 
@@ -128,7 +128,7 @@ Take `fedora-django` image and use it for `web` container, map port `8000` from 
 Let's start new project with command `django-admin`:
 
 ```
-$ docker-compose run web django-admin startproject awesome_web .
+# docker-compose run web django-admin startproject awesome_web .
 ```
 
 Since the container runs as root by default, even our project will be created as root, let's fix it:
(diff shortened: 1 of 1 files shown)


tools/docker/compose.md
```diff
@@ -8,36 +8,43 @@ section: tools
...
+3. our choice of database will be PostgreSQL (which we link to the container with sources)
+4. the whole development environment will be managed by docker-compose
+
 
 ## Prerequisites
 
+
+### Docker
+
+This guide expects that you have docker engine installed, configured and running (if your user is not able to reach docker, prefix all `docker` commands with `sudo `; for more information see section docker of this site).
```
strzibny: We might discourage people from using sudo (get back to `#` notation) since new users are not administrators by default. It's a no brainer to still use it in commands, because we would change them all in one commit (if we decide to do so), but referencing this in text seems to be a little problematic. So if it's not much to ask, can you rather mark the commands with `#`?).

TomasTomecek: sure

strzibny:
I have a suggestion for this (you mix verbs and nouns in one list):
```
In this example we'll go through a simple development environment setup:

    start a Django project
    mount sources inside a docker container (application will change as we code)
    our choice of database will be PostgreSQL (which we link to the container with sources)
    the whole development environment will be managed by docker-compose
```
I would rework it like:
```
In this example we will go through a simple Django & PostgreSQL development environment set-up managed by `docker-compose`. We will mount the application sources inside the docker container, which will allow us to change the application source code on the go, and link the database container to the application container so our Django application will be able to communicate with our PostgreSQL database.
```

Also:
```
# chown 26:26 db
```

And most importantly even though you included troubleshooting section, the tutorial still misses a step how to avoid:
```
Starting awesomeweb_db_1...
Cannot start container bg1f8cb2d227a6efa5e82d9669235430f63dadc76a5ddd4907248f1edc11490a:
    Cannot link to a non running container: /awesomeweb_db_1 AS /awesomeweb_web_run_2/awesomeweb_db_1
```
This happens _by default_ so it's necessary to write the tutorial in the way it does not happen.

Is postgres user on Fedora _always_  26? Perhaps would be nice to include a note that 26 is a user ID for postgres on Fedora, it's not clear from the text.


And at many places you still use won't, it's, we'll. Please avoid them (see https://github.com/developer-portal/content/blob/master/CONTRIBUTING.md#avoid-shorthands).

strzibny:
Btw I always test these things with clean Fedora 22 environment (using Vagrant), which is a perfect way to copy&paste the commands and see what happens on clean Fedora.

TomasTomecek:
Updated. Fixes for postgresql image were merged to Fedora-Dockerfiles, so the guide should work now properly.

strzibny:
@TomasTomecek thank you, I will merge this and go once more though it.

strzibny:
@TomasTomecek the following is still failing:

```
$ sudo docker-compose run web django-admin startproject awesome_web .
Creating awesomeweb_db_1...
Cannot start container 6c0e84112d42c63378ced4adf3889765ddc60646e4848deabe9c2f47c61c62a9: Cannot link to a non running container: /awesomeweb_db_1 AS /awesomeweb_web_run_1/awesomeweb_db_1
```
Its not a good idea to have an example that people cannot follow. Can you fix it? It probably miss some command before that would start the `web` container.

PR #84, opened 2015-11-02
Drop redundant sudo

tomhughes:
There's no need to use sudo to link a module to the working directory, and it probably isn't a good idea.


strzibny:
Thank you!

PR #90, opened 2015-11-03
docker compose: add newline before codeblock

TomasTomecek:
b/c it looks like that parser makes it and inline codeblock

https://developer.fedoraproject.org/tools/docker/compose.html#create-django-project

Don't know how to verify

### DIFF
--- a/tools/docker/compose.md
+++ b/tools/docker/compose.md
@@ -114,6 +114,7 @@ $ sudo chown -R $UID:$UID awesome_web
 ```
 
 Since we use PostgreSQL, we need change `DATABASES` variable in `settings.py`:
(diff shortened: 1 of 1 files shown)


tools/docker/compose.md
```diff
@@ -114,6 +114,7 @@ $ sudo chown -R $UID:$UID awesome_web
 ```
 
 Since we use PostgreSQL, we need change `DATABASES` variable in `settings.py`:
```
strzibny: change is a verb here right? should be `we need to` IMHO

TomasTomecek: yup

PR #95, opened 2015-11-08
Create gnome-builder.md

phracek:
Needs to be reviewed and added more templatest.

### DIFF
--- a/start/sw/gui-app/gnome-builder.md
+++ b/start/sw/gui-app/gnome-builder.md
@@ -1,0 +1,22 @@
+---
+title: GNOME Builder      # This title is shown in menus
+page: gui-app     # This is an ID of your content (select value from the list below)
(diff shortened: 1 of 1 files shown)


kvok:
It should be clear the GNOME Software means the app. There may exist a website called like that and that would be confusing.
"or find **Builder** in the GNOME Software app" would be more suitable.

phracek:
@kvok Can you please do a modification?

phracek:
The issue was done from scratch and I don't thing that it is useful for Portal at all.

felipeborges:
I would suggest placing the gnome-builder article within the tools section. It could be later linked in the gui-app section as well as in the language pages where using Builder as IDE is convenient (such as: C + Gtk+, Vala, PyGObject...)

PR #98, opened 2015-11-09
Python Django

Melancholix:
I think it needs review


phracek:
@Melancholix  Can you do that? We plan to release an update before Christmas.
It would be good to have it on portal.

Melancholix:
@phracek Of course... I'll look into it!

Melancholix:
@phracek I hope it is not too late. Review last commit...

PR #101, opened 2015-11-12
Fixed typo and small style enhancement.

voxik:


asamalik:
Thank you!

tech/languages/ruby/gems-installation.md
```diff
@@ -47,7 +47,9 @@ To install `thor` gem you therefore install `rubygem-thor` package:
 
 ```
 $ sudo dnf install rubygem-thor
+```
```
strzibny: I think we need a blank like  before/after code blocks, otherwise all will display inline. I will fix it.

PR #106, opened 2015-11-20
Distinguish between package collection and user group

voxik:
It is confusing to have word 'group' mentioned twice on single page, each time in completely different context. This is attempt to clarify this.

BTW not sure if this terminology is used on other pages, but it should be clarified/keep consistent there as well.


voxik:
I missed the colon. Could you please fix it for me?

strzibny:
I will.

Issue #99, opened 2015-11-11
Images in content


FrostyX:
It would be nice to have possibility to show images in the content. It doesn't make much sense for programming languages and CLI stuff, but in some cases it might be fine:

1. Showing image with end result as a motivation (I am working on Sphinx tutorial, and I think screenshot of some documentation would be nice)
2. Showing some GUI/Web tool (From current content it can be DevAssistant and Copr)
3. ...

pvalena:
It is possible to attach images, f.e.:

https://github.com/developer-portal/content/commit/6dbd8448556eff0626c1ba39590cb76e141c73e9

Thank you!

Issue #97, opened 2015-11-09
Python - Sphinx


FrostyX:
I think that there should be a section about Sphinx as the most commonly used documentation generator for Python.

In a case that someone is already working on this, please let me know. Otherwise I will start to work on this.

TomasTomecek:
CC @sYnfo

FrostyX:
It is already in production, so I am closing this.

PR #107, opened 2015-11-22
Add sphinx artictle

FrostyX:
This PR adds content of #97 
Can we please discuss whether you consider it useful, and what should be improved?

Also, I would like to add image there, to let users see, what result they can expect. Please see #99.

### DIFF
--- a/tech/languages/python/sphinx.md
+++ b/tech/languages/python/sphinx.md
@@ -1,0 +1,78 @@
+---
+title: Sphinx
+page: python
+order: 5
+---
+
+# Sphinx
+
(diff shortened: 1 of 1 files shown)


phracek:
@FrostyX Could you please update this PR? I don't see updates which are mentioned in What's next.
Could you please do it? Or shall I do it?

FrostyX:
@phracek As I said before

> My apologies, but I can't see a difference between actual and suggested link (same for examples link above)

phracek:
I will do fixes.

FrostyX:
Thank you for merging! :-)

PR #109, opened 2015-12-03
Section about 32-bit compilation added to c_installation.md

fkluknav:
Installing libgcc.i686 is not intuitive for people not familiar with Fedora packaging.


phracek:
Thanks.

PR #118, opened 2016-01-23
f24 feature: introduce sen, TUI for docker

TomasTomecek:
not sure what is the correct way of adding images; for now it works with github


TomasTomecek:
updated

PR #119, opened 2016-02-11
cmake: fix typos and tidy up

kragniz:


phracek:
Thanks for fixes.
CMake is really also for C projects.
MErging.

PR #123, opened 2016-04-20
Add item about PHP SIG

Conan-Kudo:
Since we have a section for PHP, we should reference the Fedora PHP SIG.


phracek:
Great thank you.

PR #124, opened 2016-04-20
Introduce pygobject article

felipeborges:
Although this article is in the tech/languages/python section, it will be referenced also in the upcoming Desktop GUI Application (gui-app) section.


phracek:
Thank you for the content. PyObject was missing.

PR #129, opened 2016-04-26
Introduce Gjs article

felipeborges:
Gjs is the GNOME Javascript binding. It will be referenced also in the upcoming Desktop GUI Application (gui-app) section.

In the metadata of this entry I implied the existence of a Javascript section. I propose it as a better way to organize the whole pletora of Javascript bindings, libraries, and frameworks (nodejs, AngularJS, Ember, gjs, ...)

### DIFF
--- a/tech/languages/gjs/gjs-installation.md
+++ b/tech/languages/gjs/gjs-installation.md
@@ -11,7 +11,7 @@ Gjs is a Javascript binding for GNOME and can be used to interact with Gtk+. It 
 To install Gjs on Fedora, run:
 
 ```
-sudo yum install gjs gjs-devel
+sudo dnf install gjs gjs-devel
 ```
 
 ## Getting started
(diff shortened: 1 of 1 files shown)


asamalik:
Hey Felipe, thanks a lot! I really like the way you wrote it - it's a simple example with some description, exactly what we want.

I think that creating a JavaScript subsection for this + move Node.js there is a good idea.

@phracek do you agree with the JavaScript subsection?

asamalik:
I'm also looking forward to the Desktop GUI Application section!

asamalik:
@pvalena the same question for you :) (do you agree with the JavaScript subsection?)

phracek:
@asamalik Definitely. Go ahead and thanks.

phracek:
Thanks for it. I am glad to have GUI application in Fedora Developer Portal.

pvalena:
Hi, sorry for the delay.
Yes, the JS section sounds good.

I especially think it's good for guiding novice developers to new techniques and utilizations, using the same language. Maybe we could have something similar for other languages?

+Thanks

PR #130, opened 2016-04-26
Introduce gtk article

felipeborges:
Although this article is in the tech/languages/c section, it will be referenced also in the upcoming Desktop GUI Application (gui-app) section.

I took the liberty to be more abstract in the C article, considering the C programmers would be a little bit more advanced the the average.


asamalik:
Very nice work (as I already said in the previous PR). Thanks a lot!

PR #135, opened 2016-05-03
Introduce Maintain and Improve section

phracek:
Signed-off-by: Petr Hracek <phracek@redhat.com>

This section introduce several tools which can be used for maintaining packages and detecting errors.
Nowadays it contains only rebase-helper and static analyses.

### DIFF
--- a/deployment/maintain/about.md
+++ b/deployment/maintain/about.md
@@ -1,0 +1,11 @@
+---
+title: Maintain and Improve
+subsection: maintain
+section: deployment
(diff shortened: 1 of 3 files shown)


phracek:
Please review it and let me know if something is wrong.
static analysis is taken from C language where it is already mentioned.

phracek:
@asamalik @pvalena : Can you please review it so we can merge it. Thanks.
Any volunter is welcome.

PR #143, opened 2016-06-22
Add JHBuild page in the tools.

meghprkh:


tools/jhbuild/about.md
```diff
@@ -0,0 +1,53 @@
...
+```
+$ git clone --depth=1 git://git.gnome.org/jhbuild
+```
+
+Cd to the source directory and build it
+```
+$ cd ~/jhbuild/
+$ ./autogen.sh
+$ make
+$ make install
```
asamalik: I am not a big fan of installing software this way. Do you think you/someone could package it as an RPM? At least in Copr.

phracek: @meghprkh Can you please update PR so that git clone is not used? Can JHBuild be packaged into Fedora?

strzibny:
Sorry but at this point we only include base Fedora packages and well-maintained Coprs.

phracek:
@meghprkh We are not able to merge it before an update of this Pull Request. Can you please update it?
Are there any problems?

meghprkh:
I can if somebody maintains a Copr. Maintaining a copr would be hard as it is updated every 2 days and so sadly it is not possible for me to maintain it. Please close the issue if you wish

phracek:
@meghprkh You can update it based on your approach. Not every two days, but monthly or weekly.
I do it for PyCharm. https://copr.fedorainfracloud.org/coprs/phracek/PyCharm/ But of course not so often.

I see a lot of benefits from JHBuild tool. Like it mentioned here https://developer.gnome.org/jhbuild/stable/introduction.html.en.

tools/jhbuild/about.md
```diff
@@ -0,0 +1,53 @@
...
+
+Clone jhbuild
+```
+$ git clone --depth=1 git://git.gnome.org/jhbuild
+```
+
+Cd to the source directory and build it
+```
+$ cd ~/jhbuild/
+$ ./autogen.sh
```
pvalena: This does not seem correct to me. It is not clear where jhbuild is cloned to. Please try to find another way of using jhbuild. Preferably without running any cloned scripts.

tools/jhbuild/about.md
pvalena: @meghprkh Yes, copr seems like a good way to compile this code. Not on a users machine. I find it is actually [quite easy](https://docs.pagure.org/copr.copr/user_documentation.html), and it can build on github hooks too. Please consider it

pvalena:
I think this needs to be packaged as RPM first(preferably Fedora or Copr).

meghprkh:
Hi, sorry for being unable to respond to it. I have not been able to find time and have been away from the GNOME ecosystem for a while now. While I don't like the idea of cloning and building it locally, it might iron out any issues before building actual packages in the gnome system since (I am assuming) JHBuild has minimal dependencies

PR #157, opened 2016-09-03
Update to Installing Rails from RubyGems.org

ianmarmour:
Currently in the latest version of Fedora with the latest versions of Ruby it's required to have rpm-build installed as well otherwise the install will fail on both nokogiri and nio4r by installing rpm-build the entire install goes off without a hitch. Adding this option will save developers time when looking to get rails installed and be able to install from RubyGems.org vs only the packaged versions.

### DIFF
--- a/tech/languages/ruby/ror-installation.md
+++ b/tech/languages/ruby/ror-installation.md
@@ -8,7 +8,7 @@ order: 4
 
 ## Installing Rails from RubyGems.org
 
-To install Ruby on Rails on Fedora as a gem, install Ruby first together with `ruby-devel`, `gcc`, `libxml2-devel` packages, and then install Rails using the `gem` command:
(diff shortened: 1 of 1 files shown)


ianmarmour:
It also appears that nodejs is a requirement for a basic install to be up and running, otherwise errors are presented by uglifier regarding finding a javascript runtime. I suggest the above commit as well as adding nodejs to the required package list from install from RubyGems.org

phracek:
@ianmarmour Why Rails needs `rpm-build`? Does it mean, for installing Rails you need to create a RPM and afterwards installing via `dnf`? Sorry for may be stupid questions. But I guess, gem can be used instead of `dnf`, right?

strzibny:
Can you please investigate the right dependency/root cause of this? Is it really `rpm-build`? Isn't it something already mentioned here: https://developer.fedoraproject.org/tech/languages/ruby/gems-installation.html?

pvalena:
The `rpm-build` was not needed for `nokogiri` last time I have checked.
Feel free to reopen with attached error message if you experience the issue again.

Closing for inactivity.

PR #159, opened 2016-09-05
Multiple Python interpreters

hroncok:
I'd also suggest to move this up a bit in the menu, but I didn't make that change so it's easier to merge in
the future without conflicts. Reordering can be done later.

### DIFF
--- a/tech/languages/python/multiple-pythons.md
+++ b/tech/languages/python/multiple-pythons.md
@@ -7,8 +7,8 @@ order: 8
 # Multiple Python Interpreters
 
 If you are working on a piece of Python software, you probably want to test it
-on multiple Python interpreters. On Fedora, that's easy, all you have to do is
-use `dnf` to install whta you need. Currently Fedora has the following Pythons
+on multiple Python interpreters. On Fedora, that's easy: all you have to do is
+use `dnf` to install what you need. Currently Fedora has the following Pythons
 ready for you in the repositories:
 
  * CPython 3.6
@@ -32,7 +32,7 @@ sudo dnf install python26  # to get CPython 2.6
 sudo dnf install pypy pypy3 jython python35  # to get more at once
 ```
 
-After that, you can run an interactive console or your script with let's say
+After that, you can run an interactive console or your script with, let's say,
 CPython 3.4:
 
 ```
@@ -58,7 +58,7 @@ you'll automatically get all the CPythons and PyPys:
 sudo dnf install tox
 ```
 
-If your are not yet familiar with tox, don't worry, this short example will show
+If your are not yet familiar with tox, don't worry. This short example will show
 you how to start.
 
 Let's create a directory and a simple Python file in it that will say something nice:
@@ -69,7 +69,7 @@ print('Fedora is the best OS for Python developers', end='\n\n')
 ```
 
 Now we'll test if it works with all the Pythons, with tox.
-We'll create a simple configuration file for tox, called `tox.ini` in the same
+We'll create a simple configuration file for tox, `tox.ini`, in the same
 directory:
 
 ```
@@ -81,9 +81,9 @@ commands=python say.py
 ```
 
 The `envlist` directive defines the list of Pythons to test on.
-Normally, tox assumes you are testing a project with it's own `setup.py`, for
-the simlicity of this demo, we are not suing it, so we tell it to tox via the
-`skipsdist` option.
+Normally, tox assumes you are testing a project with its own `setup.py`. For
+the simlicity of this demo, we are not using it, and we need to tell this to
+tox via the `skipsdist` option.
 Finally the `commands` in `[testenv]` section tells tox what commands to run
 for the test, normally that would be `python setup.py test`, `py.test` or
 similar.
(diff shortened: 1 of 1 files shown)


tech/languages/python/multiple-pythons.md
```diff
@@ -0,0 +1,70 @@
...
+
+```bash
+python3.4 -m ensurepip --user
+python3.4 -m pip install --user <package>
+```
+
+Or even create virtualenvs:
+
+```bash
+python3.4 -m venv <dir>
```
encukou: The `venv` module was added in 3.3, for 2.6 and 2.7 you need to use python-virtualenv.
MicroPython has [upip](https://pypi.python.org/pypi/micropython-upip/) -- `micropython -m upip`, works with caveats but installs to `~/.micropython/lib`.
Not sure what the procedure is with PyPy and Jython.

hroncok: Would this be better?

    virtualenv --python /usr/bin/python3.4

Or should I specify the command for each (group of) interpeter(s)?

tech/languages/python/multiple-pythons.md
```diff
@@ -0,0 +1,70 @@
...
+CPython 3.4:
+
+```bash
+python3.4
+```
+
+Or get pip and use it to install packages:
+
+```bash
+python3.4 -m ensurepip --user
```
encukou: Ensurepip was added in 3.4.

tech/languages/python/multiple-pythons.md
encukou: I think "installing packages" should be its own section that would explain stuff is not packaged in Fedora and give specific instructions for each group of interpreters.

tech/languages/python/multiple-pythons.md
```diff
@@ -0,0 +1,70 @@
...
+```bash
+python3.4 -m venv <dir>
+```
+
+Note that most of the CPython versions, other than those provided by the
+`python2` and `python3` packages, are only provided for developer's convenience
+so you can run your test suites. Critical security bugs might not be always
+fixed and those interprets are not intended for production.
+
+## Getting it and running it all with tox
```
encukou: This is the main use case. I'd put it after the list of available interpreters, but before mentioning how to install things manually.
IMO the section needs a short introduction on what tox is.

tech/languages/python/multiple-pythons.md
encukou: Perhaps just explaining virtualenv/venv would be enough here. Going with `pip --user` first is not good for experimenting – it pollutes the user namespace and adds stuff to $PATH that should only be used for testing.

hroncok: Ok, so only explain creating virtualenvs and installing stuff inside?

encukou: Right, with a mention that there are no Fedora packages for the alternative Pythons.

hroncok:
Amended

hroncok:
Could i please get some feedback from the repository owners as well?

Also, is it later better to indicate that some parts are let's say Fedora 25+ only, or wait with the merge until F24 is EOL? I prefer the first option, but not sure what's the common practice here.

tech/languages/python/multiple-pythons.md
```diff
@@ -0,0 +1,202 @@
+---
+title: Multiple Pythons
+subsection: python
+order: 8
+---
+
+# Multiple Python Interpreters
```
strzibny: Just keep the headline as the others, e.g. `Multiple Python interpreters`

strzibny:
So far I believe we only published things for current Fedora, perhaps with notes mentioning previous actively maintained versions. 

@phracek @asamalik I suggest create a fedora-25 label or future or something.

hroncok:
In this case: Multiple python interpreters will be in F23, F24, etc... the only thing thet will be F25+ only is tox bringing them in.

strzibny:
Then we can merge the part without tox, and tox should be kept open....

tech/languages/python/multiple-pythons.md
```diff
@@ -0,0 +1,202 @@
...
+## Getting it and running it all with tox
+
+[Tox](https://tox.readthedocs.io/) is tool that helps you test your Python code
+on multiple Pythons. If you install it on Fedora via the dnf package manager,
+you'll automatically get all the CPythons and PyPys:
+
+```bash
+sudo dnf install tox
+```
+
```
hroncok: What about here:

> This will only install the Python interpreters on Fedora 25 or newer, on older Fedoras, you need to install the packages specifically.

And later remove that note once F24 EoLs.

strzibny: It's probably true that this is IMHO acceptable.

phracek:
Fedora Developer portal has a focus to current "rawhide" version. Now it is Fedora 25 which will be released soon.

tech/languages/python/multiple-pythons.md
```diff
@@ -0,0 +1,202 @@
...
+If your are not yet familiar with tox, don't worry. This short example will show
+you how to start.
+
+Let's create a directory and a simple Python file in it that will say something nice:
+
+```python
```
phracek: python should be on new line

phracek:
But seems to be fine.

tech/languages/python/multiple-pythons.md
```diff
@@ -0,0 +1,202 @@
...
+ * PyPy 3
+ * Jython
+ * MicroPython
+
+Quite a nest, isn't it?
+You can install them like this:
+
+```bash
+sudo dnf install python33  # to get CPython 3.3
+sudo dnf install python34  # to get CPython 3.4
```
hroncok: As said: "The text refers to features not yet available! Do not merge yet, but please provide feedback for the content."

tech/languages/python/multiple-pythons.md
hroncok: It's Markdown Syntax, not a text that should be seen.

hroncok:
This is now true. 

```
$ sudo dnf install tox
Dependencies resolved.
=====================================================================================================
 Package                     Arch            Version                  Repository                Size
=====================================================================================================
Installing:
 pypy                        x86_64          5.4.0-1.fc25             fedora                    26 k
 pypy-libs                   x86_64          5.4.0-1.fc25             fedora                    17 M
 pypy3                       x86_64          2.4.0-6.fc25             fedora                   6.7 M
 pypy3-libs                  x86_64          2.4.0-6.fc25             fedora                   5.0 M
 python-rpm-macros           noarch          3-10.fc25                fedora                   9.1 k
 python-srpm-macros          noarch          3-10.fc25                fedora                   8.4 k
 python26                    x86_64          2.6.9-2.fc25             fedora                   9.8 M
 python3-devel               x86_64          3.5.2-4.fc25             fedora                   204 k
 python3-pluggy              noarch          0.3.1-5.fc25             fedora                    28 k
 python3-py                  noarch          1.4.31-3.fc24            fedora                   467 k
 python3-rpm-macros          noarch          3-10.fc25                fedora                   8.5 k
 python3-tox                 noarch          2.3.1-4.fc25             fedora                    89 k
 python3-virtualenv          noarch          15.0.3-1.fc25            fedora                   1.7 M
 python33                    x86_64          3.3.6-1.fc25             fedora                    12 M
 python34                    x86_64          3.4.5-1.fc25             fedora                    15 M
 tcl                         x86_64          1:8.6.6-1.fc25           fedora                   2.2 M
 tix                         x86_64          1:8.4.3-17.fc24          fedora                   259 k
 tk                          x86_64          1:8.6.6-1.fc25           fedora                   1.6 M

Transaction Summary
=====================================================================================================
Install  18 Packages

Total download size: 73 M
Installed size: 346 M
```

strzibny:
Fedora 25 is beta already, I think it's safe to merge. Thanks!

hroncok:
Thanks. When/how gets this deployed to https://developer.fedoraproject.org/ ?

PR #160, opened 2016-09-23
Page about Micropython

frenzymadness:
Hello.

We are working on MicroPython related stuff in Fedora and it would be nice to have MicroPython page on developer portal - so here is our PR with a new page describing MicroPython and related tools.

Lumir Balhar, Python maintenance team, Red Hat


phracek:
Sounds good to me. @jhornice Can you please review it from English point of view?

frenzymadness:
Thanks for merge. When will it be available on the live portal?

phracek:
It is already available on portal. https://developer.fedoraproject.org/tech/languages/python/micropython.html

PR #167, opened 2016-11-10
Fixed copr API URL

ssbarnea:
The old url was not the correct one, if one would try to login to get his API key, he would be redirected to the homepage of the new site. In fact I have the impression that the entire URL should be using a HTTP redirect to the new domain.


phracek:
@asamalik Can you please confirm it?

asamalik:
Thanks!

PR #169, opened 2016-11-26
fix laravel command not found

mmhyamin:
Issue: http://stackoverflow.com/a/27915343/3060426

### DIFF
--- a/start/sw/web-app/laravel5.md
+++ b/start/sw/web-app/laravel5.md
@@ -58,10 +58,10 @@ $ composer -V
 $ composer global require "laravel/installer"
 # Place the ~/.config/composer/vendor/bin directory in your PATH so the laravel 
 # executable can be located by your system.
-$ echo 'export PATH="$PATH:$HOME/.config/composer/vendor/bin"' >> ~/.bash_profile
+$ echo 'export PATH="$PATH:$HOME/.config/composer/vendor/bin"' >> ~/.bashrc
 ```
 
-Now you can create a laravel project with the commands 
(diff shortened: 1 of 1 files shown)


pvalena:
My .bash_profile already contains `export PATH` and .bashrc seems like a strange place for path.
Could you elaborate on why is this an issue?

start/sw/web-app/laravel5.md
```diff
@@ -58,10 +58,10 @@ $ composer -V
 $ composer global require "laravel/installer"
 # Place the ~/.config/composer/vendor/bin directory in your PATH so the laravel 
 # executable can be located by your system.
-$ echo 'export PATH="$PATH:$HOME/.config/composer/vendor/bin"' >> ~/.bash_profile
```
pvalena: You do not need the `export` here. PATH should be already exported on all UNIX systems.

For immediate change in current terminal you could run the command `source !$` instead, which sources ~/.bash_profile.

mmhyamin: @pvalena could you please take a look at the [stackoverflow](http://stackoverflow.com/questions/26376516/laravel-php-command-not-found) issue?

pvalena: @mmhyamin I sure already did, but I think the answer you have sent is not the correct one (FTR it's not the accepted one) in this case.
Could you check my proposed solution?
Please clarify the issue you're having if it is not resolved by it (`source !$`, right after the `PATH=". . .`).

pvalena: We'll resolve the issue later.

PR #174, opened 2016-12-23
gui is includeded in devassistant main pkg

bithakr:
https://admin.fedoraproject.org/pkgdb/package/rpms/devassistant/ - "This package will install both the command-line and graphical interfaces."


sferich888:
Then why are we bundling 2 packages?

pvalena:
Already solved in https://github.com/developer-portal/content/pull/178.

Thanks!

PR #178, opened 2017-01-20
Improved android-studio setup documentation :v:

GabLeRoux:
* `my_test` typo for consistency
* I wasn't sure at first about the `--name <your_project_name>` param,
  but it's mandatory for the command. I didn't need to create a project,
  I only wanted Android Studio for some existing projects, but it can definitely be deleted after.
* I added official instructions as we sometimes fall on devassistant
  before the actual original documentation
* I added troubleshooting part concerning missing JDK Setup


phracek:
Thx for the update. Make sense to me.

pvalena:
Just one question: does it not work win openjdk? There is Oracle JDK referenced.

PR #179, opened 2017-02-01
vagrant: fix title of atlas page

dustymabe:


pvalena:
Please note that all changes are reflected on testing server[1] in ~10 minutes time after they're merged.

http://developer.fedorainfracloud.org/tools/vagrant/about.html

PR #186, opened 2017-03-23
Virtualenv is possible with Jython now

hroncok:


pvalena:
LGTM

pvalena:
You can check your changes here:

http://developer.fedorainfracloud.org/tech/languages/python/multiple-pythons.html

PR #189, opened 2017-04-05
venv creation syntax update

offthewoll:
As of Python 3.6, pyvenv is deprecated in favor of python3 -m venv (see https://docs.python.org/3/library/venv.html#module-venv).


pvalena:
@hroncok, is this change fine for you? Thanks!

hroncok:
Sure

pvalena:
Great, thanks!

PR #207, opened 2017-07-05
Replace fedorahosted.org links with updated URLs

ferdnyc:
fedorahosted.org is gone, and the COPR documentation now lives on pagure.io. I chased down equivalent URLs for all of the documentation links that had gone rotten.


ferdnyc:
(Note that after I filed this request, I got curious about the state of the rest of the pages, so I cloned the entire content repo and grepped it for instances of 'fedorahosted.org'. Happily, the three I've corrected here were the only ones found.)

pvalena:
LGTM, thanks!

PR #211, opened 2017-07-12
Remove nose from Scientific Python

hroncok:
* nose has nothing to do with science
 * nose is dead, it says the following on the website:

> Nose has been in maintenance mode for the past several years
> and will likely cease without a new person/team to take over
> maintainership. New projects should consider using Nose2,
> py.test, or just plain unittest/unittest2.


pvalena:
Thanks!

PR #216, opened 2017-07-19
Include tip on overcoming UDP protocol error with Vagrant & NFS

strzibny:


tools/vagrant/vagrant-nfs.md
```diff
@@ -39,6 +39,20 @@ Vagrant.configure("2") do |config|
 end
 ```
 
+One of the common errors when using NFS this way might be that your NFS version does not support UDP protocol which gets added by Vagrant as a default. 
+
+The error says `mount.nfs: requested NFS version or transport protocol is not supported`.
```
pvalena: How about rewording to something simillar to:
```
In that case the error message is ` ... `. You can work around this by disabling UDP with `nfs_udp` option: 
[ . . . ]
```

tools/vagrant/vagrant-nfs.md
```diff
@@ -39,6 +39,20 @@ Vagrant.configure("2") do |config|
...
+
+In that case it's enough to disable UDP with `nfs_udp` option:
+
+```
+Vagrant.configure("2") do |config|
+  # ...
+
+  config.vm.synced_folder ".", "/vagrant", type: "nfs", nfs_udp: false
```
pvalena: > config.vm.synced_folder ".", "/vagrant", type: "nfs", nfs_udp: false

Would not this line simply suffice? Is it worth repeating `Vagrant.configure("2") do |config|`?
WDYT?
￼

strzibny:
I updated it as you suggested.

pvalena:
Great, thanks!

PR #220, opened 2017-08-18
url was not working

christianbueno1:
this git@github.com:officert/mongotron.git wasn't working


pvalena:
When you go to the [web page](https://github.com/officert/mongotron) of the repository and click `Clone or download` you will get

`git@github.com:officert/mongotron.git`

which is exactly what is on the page now. I do not know why it does not work for you, but as it's the default on github I'm keeping it like this.

Thanks!

PR #223, opened 2017-10-08
python-installation: fix typos, improve grammar

whydoubt:


tech/languages/python/python-installation.md
```diff
@@ -32,17 +32,17 @@ print('Hello World!')
...
+To run a program written in Python, type `python3` followed by the path and name of the program. Like this:
 
 ```bash
 $ python3 example.py
 ```
 
 ## Using virtualenv
 
-When you work at some project it is good to keep it inside a virtual environment. It will keep the dependencies you need at one place and you no dot have to worry about different projects which needs different version of the same module.
+When you work with a project, it is good to keep it inside a virtual environment. It will keep the dependencies you need in one place and you do not have to worry about different projects which need different versions of the same module.
```
pvalena: Shouldn't this be 'on' a project?
(I know it implies you're actually working on it, but isn't that the case?)
`with` does not fit well there with me.

tech/languages/python/python-installation.md
```diff
@@ -32,17 +32,17 @@ print('Hello World!')
...
 
-Let's create a virtual environment called `project_venv` which will contain Python and pip which you can use to install project's dependencies.
+Let's create a virtual environment called `project_venv` which will contain Python and pip, which you can use to install a project's dependencies.
```
pvalena: The sentence IMO contains too many `which`'s. I think the sentence ought be split in two, to clarify the usage:

`You can use pip to install project's dependencies.`

And, also
`a project's dependencies.`
is not supposed to have `a` you've added AFAIK.

tech/languages/python/python-installation.md
whydoubt: 'with' and 'on' have slightly different implications, but either works in this context.  I am fine with changing this to 'on'

tech/languages/python/python-installation.md
whydoubt: Yes, 'which' appearing twice is a bit awkward.  I agree the sentence should be split.  It also removes some ambiguity.

In this context, the word 'project' is a singular common count noun, and as such needs a determiner, such as 'a'.  Some alternatives would be "install project dependencies" where 'project' is a modifier noun, or "install projects' dependencies" where it is made plural possessive.

tech/languages/python/python-installation.md
pvalena: Sure :) Thanks.

tech/languages/python/python-installation.md
pvalena: Thanks.

Interesting :) I did not know that. I'll leave it up to you then (current is fine).

pvalena:
Please ammend the current commit with the changes.

pvalena:
Thanks a lot for your PR!

pvalena:
Thanks!

PR #224, opened 2017-10-11
Update Eclipse/aCute C# development

RheaAyase:


tech/languages/csharp/csharp-ide.md
```diff
@@ -22,29 +22,30 @@ JetBrains Rider is full featured IDE for all things C#, and even F# on .NET Core
...
+_.NET Core and Mono_
 
-[monodevelop.com](http://www.monodevelop.com)
+You can use the [aCute plugin](https://marketplace.eclipse.org/content/acute-c-edition-eclipse-ide-experimental) on top of the established and popular [Eclipse IDE](/tools/eclipse/about.html). This C# extension for Eclipse IDE relies on [OmniSharp](http://www.omnisharp.net) and [Language Server Protocol](https://github.com/Microsoft/language-server-protocol). It comes with complete edition features for C# in the Eclipse IDE. Relying on the Eclipse IDE provides all typical features of a good IDE (Polyglot development, SCM, deployment...) and give access to a rich ecosystem of extensions for about all existing technologies.
 
 ```
-$ sudo dnf install monodevelop
+$ sudo dnf install eclipse eclipse-mpc
 ```
+then install the aCute extension from [Eclipse Marketplace](https://marketplace.eclipse.org/content/acute-c-edition-eclipse-ide-experimental)
```
pvalena: `How?`
Isn't there supposed to be hint- something like 'follow this guide' perhaps?

pvalena: I've just noticed you've only changed the order of blocks. However, since you're updating it, could we enhance it too?

tech/languages/csharp/csharp-ide.md
```diff
@@ -22,29 +22,30 @@ JetBrains Rider is full featured IDE for all things C#, and even F# on .NET Core
...
 ```
-$ sudo dnf install eclipse eclipse-mpc
+$ sudo dnf install monodevelop
 ```
-then install the aCute extension from [Eclipse Marketplace](https://marketplace.eclipse.org/content/acute-c-edition-eclipse-ide-experimental)
 
 ## Summary
 
-What would we recommend? JetBrains Rider is the best C# IDE, however it is not open source. VS Code is great for .NET Core and MonoDevelop for Mono. C# in Eclipse IDE is work in progress, and supports only Mono for now; it can already be useful if you're already an user of Eclipse IDE with any other programming language.
+What would we recommend? JetBrains Rider is the best C# IDE, however it is not open source. VS Code is great for .NET Core and MonoDevelop for Mono. C# in Eclipse IDE is still young and not very friendly to former Windows developer; but it can be useful if you're already an Eclipse IDE user.
```
pvalena: Please link to those `JetBrains` and `VS Code` (and I'd rather not use abbrev., unless they're official/marketed as such).

`'Windows developer; but ` - I'd prefer `,` instead of `;` here.

pvalena: `What would we recommend?`
I do not think you're answering the question in this block. Could it be formulated differently?
Or possibly do, instead of `summary`, something like:

## What IDE to use?
 - **JetBrains Rider** - blabla...
 - **VS Code** - blabla...
 . . .

pvalena:
Btw. I can see only the blocks rearranged.. or is there some edited content I've missed?

tech/languages/csharp/csharp-ide.md
RheaAyase: I don't know how, I have **failed** to actually use eclipse with c#. It imho sucks =)
On the plugin website they say "drag this thing into eclipse ide and it will install" which doesn't work for me. I had to find it in the eclipse ide's search which was weirdly hidden, and even then, I was unable to edit any C# files or netcore projects.

That said, that's just me, it's supposed to be "an awesome ide that actually works," so I'm not gonna argue with that or tell people to not try on their own.

pvalena:
Ok, I'd say the `Marketplace` name tells us the whole story:
it's not packaged (or distributed) by or specifically to Fedora (in RPMs or Copr), but rather IDE-specific extension support.

Thanks for the PR!

tech/languages/csharp/csharp-ide.md
RheaAyase: It's supposed to be a summary, the list of IDEs with description is above and it would seem weird to have two lists of IDE's..

RheaAyase:
> Ok, I'd say the Marketplace name tells us the whole story:
>it's not packaged (or distributed) by or specifically to Fedora (in RPMs or Copr), but rather IDE-specific extension support.

OH that's what you mean... yeah it's not packaged anywhere, it's just a website with plugins and you like... "drag it in" and it's supposed to install, but as I said, that didn't work for me (probably cause of KDE instead of GNOME or something like that) so it wasn't as straight forward as it's meant to be.

tech/languages/csharp/csharp-ide.md
pvalena: Well, my point is that it's just formated in a way that summary should be, but the content is just serialized list of IDEs.

RheaAyase: I don't think that we should really say "use this," so I'm trying to make it at least somewhat objective comparison with some pro's and con's. I don't know how else to achieve that...

pvalena: It's fine :) it was just an idea.
If you think of some change, please open new PR. 

I've got one additional question, though: is `VS Code` official name/short? Because otherwise it should be fixed.

RheaAyase: Well it's Visual Studio Code but yes, VS Code is kinda official name of the project, maybe not marketing name though. They're not very consistent with the different names... Their official repo is called `/etc/yum.repos.d/vscode.repo` then you have `dnf install code` and the official name is Visual Studio Code. It's kinda messy, but everyone is using VS Code so I think that it's fine.

https://code.visualstudio.com
https://github.com/Microsoft/vscode

tech/languages/csharp/csharp-ide.md
pvalena: Come to think of it, maybe we should at least mention it does not have to work, or there might me some difficulties.

tech/languages/csharp/csharp-ide.md
pvalena: `kinda`
I'll fix it at least for title....

PR #229, opened 2017-10-23
*csharp IDE: Removed the bug notice as it's hacked now

RheaAyase:


tech/languages/csharp/csharp-ide.md
```diff
@@ -18,8 +18,6 @@ JetBrains Rider is full featured IDE for all things C#, and even F# on .NET Core
 
 _.NET Core Only_
 
-* VS Code will not work with [our .NET Core packages](/tech/languages/csharp/dotnet-installation.html) due to an [upstream issue](https://github.com/dotnet/source-build/issues/125) affecting OmniSharp. You can still use it with [Microsoft's dotnet packages](https://www.microsoft.com/net/core#linuxfedora).
-
 Visual Studio Code (aka VS Code) is plugin-based IDE that by itself won't just run C#. You have to install the C# plugin, and NuGet Package Manager.
 
 Download the Visual Studio Code at [code.visualstudio.com](https://code.visualstudio.com) and the C# Extension at [marketplace.visualstudio.com](https://marketplace.visualstudio.com/items?itemName=ms-vscode.csharp)
```
pvalena: `Download the Visual Studio Code at [code.visualstudio.com](https://code.visualstudio.com) `

Seeing this- is it a complete solution? Is there a guide and everything to make it work on Fedora?

RheaAyase: It's a... "download .rpm here" ...I don't think that we need to explain how to install an rpm? o.o

pvalena: > It's a... "download .rpm here" ...I don't think that we need to explain how to install an rpm? o.o

Nope, I [checked it](https://code.visualstudio.com/docs/setup/linux#_rhel-fedora-and-centos-based-distributions) and It's a full guide. Could we, possibly, point to it directly?
Additionally we could link an [alternate installation](https://code.visualstudio.com/docs/setup/linux#_installing-rpm-package-manually).

Anyways; AFAICT it has an error:
```
dnf check-update
# and
yum check-update
```
is missing `sudo` (`dnf` has per-user database).

```
sudo rpm --import https://packages.microsoft.com/keys/microsoft.asc
sudo sh -c 'echo -e "[code]\nname=Visual Studio Code\nbaseurl=https://packages.microsoft.com/yumrepos/vscode\nenabled=1\ngpgcheck=1\ngpgkey=https://packages.microsoft.com/keys/microsoft.asc" > /etc/yum.repos.d/vscode.repo'
```
is not very pretty, but should work, I hope.

PR #238, opened 2017-12-20
Added page about Hugo

yotamN:


pvalena:
Please squash the second commit into second one (`fixup`). Do not hesitate to ask if you do not know how.

Could you also add reference page `Go Frameworks` or similar to [go section](https://developer.fedoraproject.org/tech/languages/go/go-installation.html)?

Thanks for the PR!

start/sw/web-app/hugo.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+To use the theme you need to add the theme parameter to the config.toml file
+
+## Running Hugo built in server
+
+Hugo comes with a built in web server to use if you don't want to install a web server. To run the server, you need to type:
+
+```
+hugo server
```
pvalena: You're missing a `$` here.

pvalena:
@yotamN PTAL on my comment.

yotamN:
I think I managed to squash the commits.
Abount the reference I'm not sure if Go Frameworks is the correct name for it. It's called a static site generator but Hugo is the only popular one in Go that I know about.

pvalena:
LGTM.
Thanks!

Issue #237, opened 2017-12-20
move pages to proper subsections


yotamN:
When there are a lot of pages on one subject it will be easier to find what you want with sub pages, for example in PHP there could be frameworks group with Symfony, Zend, CodeIgniter, Laravel etc.

Is it possible in the current template or does it require a change?

pvalena:
For reference: http://developer-portal.github.io/doc/structure.html

`Start a project > Web Application > Laravel ... `
and 
`Tools > Web Server`
and
`Languages > PHP > Web Frameworks`

is what makes sense IMHO. Needs a bit of fixing though.

Thans for finding this!

pvalena:
I think it should say
`Web Frameworks`
specifically.

http://developer.fedorainfracloud.org/tech/languages/php/php-frameworks.html

Also, no need to prefix the file with `php-` when it already is in php subsection.

pvalena:
I've just learned that we can set permalinks to the duplicated pages. F.e.:
```
permalink: /tech/languages/php/laravel5.html

```

yotamN:
@pvalena In web applications for example you could have HTTP servers, Static Site Generators and Web Frameworks as 3 subsections or probably sub subsections because  web applications is the sub section

pvalena:
That seems to be overcomplication to me...
One could simply have a page `HTTP servers` that has a hyperlink list to other pages.

Like here: http://developer.fedorainfracloud.org/tech/languages/php/php-frameworks.html

pvalena:
Feel free to submit a PR with any `cathegory` type of page you thing is missing.

yotamN:
The example you gave of `Languages > PHP > Web Frameworks` could stay the same but instead of a page that contains link it will be an expandable list in the sidebar. The content will remain the same, just a bit more organized

pvalena:
On top of that, you can actually prefix the page name in the pages list like so:
```
Frameworks
 - Zend
 - Laravel5
 - [ . . . ]
```

yotamN:
Level of complexity to the site development or the content? It's true that the site will need some work to support it but after the initial work I don't think it will be that much complexity to the content writers

jackorp:
Created new issue regarding this: #386

PR #239, opened 2017-12-22
Moved all php web frameworks to web app

yotamN:
First part of #237 

I will work on the second part (creating a web servers subsection) next week.


pvalena:
Thanks for the PR!

PR #241, opened 2017-12-23
Update rpm and copr-cli pages

galoget:


deployment/rpm/about.md
```diff
@@ -128,7 +128,7 @@ $ rpmbuild -bs hello.spec
 Then you can run:
 
 ```bash
-$ mock -r epel-7-i386.cfg hello.src.rpm
+$ mock -r /etc/mock/epel-6-i386.cfg ~/rpmbuild/SRPMS/hello-2.10-1.fc27.src.rpm
```
pvalena: Why `epel6`? Also, why `i386` and not `x86_64`? (I just wonder if there are some reasons behind it?)
Also I'd say it's really not needed to use full path.

deployment/rpm/about.md
```diff
@@ -23,10 +23,10 @@ Biggest advantages of packaging your application is:
 Run:
 
 ```bash
-$ sudo dnf install fedora-packager rpmdevtools
+$ sudo dnf install fedora-packager rpmdevtools gcc
```
pvalena: I wonder why would you need gcc when you're building in mock if I'm not mistaken.

pvalena:
Thank you for your PR!

pvalena:
@galoget, any news on this?

deployment/rpm/about.md
Conan-Kudo: This should absolutely be for latest fedora and el, not el6...

deployment/rpm/about.md
ferdnyc: You're not _only_ building in mock. § Quick Start Guide does end with a suggestion to run `rpmbuild -ba hello.spec`, and GNU Hello is a C program. 

Technically you'll need other build tools besides `gcc` to compile GNU Hello, tho: `make`, `autoconf`, etc. I don't _believe_ the three packages listed will necessarily pull all of those in, if they're not already installed.

pvalena:
`This branch has conflicts that must be resolved
`

deployment/rpm/about.md
pvalena: > for latest fedora and el, not el6...

Actually, as long as this works *on Fedora* it does not matter.

deployment/rpm/about.md
pvalena: Ok, let's go with that, then.
Thanks!

deployment/rpm/about.md
pvalena: Ok, let's assume the Contributor has tested this and the values do not matter in the end as it's just an example.

pvalena:
I've resolved those. I hope I did not miss anything.
Thanks!

PR #242, opened 2017-12-31
Fix Django page in `Web Application` subsection

thunderoy:
corrected the commands for installing virtualenvwrapper in python 2 and 3 and some formatting

### DIFF
--- a/start/sw/web-app/django.md
+++ b/start/sw/web-app/django.md
@@ -11,102 +11,102 @@ Django is one of the most popular Python Web Frameworks. It has very useful web 
 Because you will need to work with Python packages, write this command to your command line.
 
 ```
-sudo dnf install python3-pip
+$ sudo dnf install python3-pip
 ```
 
-# Get started
+## Get started
 
-## How to install virtualenv
+### How to install virtualenv
 
 It's recommended to keep your project inside virtual environment. But you can skip this step if you want. 
 Enter this command to your command line for installation:
 
 ```
-sudo dnf install python3-virtualenv
+$ sudo dnf install python3-virtualenv
 ```
 
 
 Get inside your working directory or create new where you want to store your virtual environment:
 
 ```
-mkdir working-dir
-cd working-dir
+$ mkdir working-dir
+$ cd working-dir
 ```
 
 Now we will create virtual environment called venv inside this directory using following command:
 
 ```
-virtualenv venv
+$ virtualenv venv
 ```
 
 When it's done, we will activate our virtualenv. Now we will use following commands:
 
 ```
-cd venv
-source bin/activate
+$ cd venv
+$ source bin/activate
 ```
 
 When it's activated we will see indicator of same name as our virtualenv. 
 Now we will check if we really have empty environment:
 
 ```
-pip freeze
+$ pip freeze
 ```
 
 It will return nothing.
 
-## How to install Django
+### How to install Django
 
 Now if you installed `virtualenv` do not leave it, we will install Django now. If you installed `virtualenv` use this command 
 inside your working directory with activated `virtualenv`:
 
 ```
-pip3 install django
+$ pip3 install django
 ```
 
 If you are following a book or a tutorial, you might want to specify the same exact version like this
 
 ```
-pip3 install django==1.10
+$ pip3 install django==1.10
 ```
 
 
 Of course you can use the version that is prepared by Fedora community by typing:
 
 ```
-sudo dnf install python3-django
+$ sudo dnf install python3-django
 ```
 
 and when it's finished use following command to start your first project called mysite:
 
 ```
-django-admin startproject mysite
+$ django-admin startproject mysite
 ```
 
 Now if you will check your directory using ls you will see new directory called mysite. We 
 will go inside that directory and we will run our project by following commands:
 
 ```
-cd mysite/
-python3 manage.py runserver
+$ cd mysite/
+$ python3 manage.py runserver
 ```
 
 You will see message with information about your development server. When you insert URL into your browser, 
 you will see welcome page (default URL for development server is http://127.0.0.1:8000/).
 And you successfully started your first Django project.
 
-## Switching Virtualenvs
+### Switching Virtualenvs
 
 You can use `virtualenvwrapper` after installing it with usual command
 
 ```
-sudo dnf install python3-virtualenvwrapper
+$ sudo dnf install python3-virtualenvwrapper
 ```
 
 ```
-mkvirtualenv env1
-mkvirtualenv env2
-workon env1
-workon env2
+$ mkvirtualenv env1
+$ mkvirtualenv env2
+$ workon env1
+$ workon env2
 ```
 
--- a/start/sw/web-app/django.md
+++ b/start/sw/web-app/django.md
@@ -10,14 +10,6 @@ Django is a high-level Python Web framework that encourages rapid development an
 Django is one of the most popular Python Web Frameworks. It has very useful web page. Visit it by clicking on [this link](https://www.djangoproject.com/). 
 Because you will need to work with Python packages, write this command to your command line.
 
-For Python 2:
-
-```
-sudo dnf install python-pip
-```
-
-or Python 3:
-
 ```
 sudo dnf install python3-pip
 ```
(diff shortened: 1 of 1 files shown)


hroncok:
Thanks to this PR I've just find out there's a Django page here. Maybe this is an unnecessary duplication with https://github.com/developer-portal/content/blob/master/tech/languages/python/django-installation.md ?

Also, Django is no longer supported on Python 2 (except the LTS version), so if we decide to keep this page, we might need to chop off the Legacy Python parts. Even if we decide to keep them, I'm not happy to see Python 2 mentioned first.

@thunderoy Would you like to do that, or should I wait for this to get merged and submit a next PR?

thunderoy:
@hroncok I think it would be better to keep the [django.md](https://github.com/developer-portal/content/blob/master/start/sw/web-app/django.md) file as it contain some extra information other than the common parts i.e. about installing pip, virtualenv and virtualenvwrapper and make the [django-installation.md](https://github.com/developer-portal/content/blob/master/tech/languages/python/django-installation.md) to redirect to django.md.

Since, python 2 is still supported in the django's LTS versions, we should keep the python 2 parts. But one thing can be done as said by you is to make it after python 3, so I will update this PR. Thanks :)

hroncok:
Still, if this is a quick how to set up a django project guide, I'd recommend not bothering newcomers with Python 2. Note that there will be no python2-django in Fedora 28+.

The virtualenv parts are described in https://github.com/developer-portal/content/blob/master/tech/languages/python/python-installation.md and https://github.com/developer-portal/content/blob/master/tech/languages/python/multiple-pythons.md

thunderoy:
@hroncok I agree on not bothering newcomers with python 2, so I will update the PR removing the python 2 parts but the above two links you have mentioned does not discuss about how to **install** virtualenv, pip and virtualenvwrapper.

start/sw/web-app/django.md
```diff
@@ -126,8 +98,9 @@ And you successfully started your first Django project.
 ## Switching Virtualenvs
 
 You can use `virtualenvwrapper` after installing it with usual command
+
 ```
-dnf install python-virtualenvwrapper
+sudo dnf install python3-virtualenvwrapper
```
pvalena: Please prepend all commands with `$`.
See our [contribuition guide](http://developer-portal.github.io/contributing/write-general-writing-hints.html).

pvalena:
@thunderoy Thanks for the PR, @hroncok thanks for the review!
Please update the remaining bits.

pvalena:
Could you also please remove ` # Get started` (or 'demote' it with `##`), because it's a first level title and there should be only one at the top of the page.

thunderoy:
@pvalena done the needful.

start/sw/web-app/django.md
```diff
@@ -10,130 +10,103 @@ Django is a high-level Python Web framework that encourages rapid development an
...
 ```
-sudo dnf install python3-pip
+$ sudo dnf install python3-pip
 ```
 
-# Get started
+## Get started
 
-## How to install virtualenv
+### How to install virtualenv
```
pvalena: Please, if possible, link to an our already existing content / section and do not duplicate content.

f.e.: Install [virtualenv](https://developer.fedoraproject.org/tech/languages/python/python-installation.html#using-virtualenv) - `/tech/languages/python/python-installation.html#using-virtualenv`

start/sw/web-app/django.md
```diff
@@ -10,130 +10,103 @@ Django is a high-level Python Web framework that encourages rapid development an
...
-
-```
-sudo dnf install python-pip
-```
-
-or Python 3:
-
 ```
-sudo dnf install python3-pip
+$ sudo dnf install python3-pip
```
pvalena: Please, if possible, link to an our already existing content / section and do not duplicate content.

See the example in another comment. 

Also, from the PyPi section I've understood that pip3 is pre-installed on Fedora. Im I right?

start/sw/web-app/django.md
```diff
@@ -10,130 +10,103 @@ Django is a high-level Python Web framework that encourages rapid development an
...
 ```
 
 
 Get inside your working directory or create new where you want to store your virtual environment:
 
 ```
-mkdir working-dir
-cd working-dir
+$ mkdir working-dir
+$ cd working-dir
```
pvalena: These steps are missing from the [python page](https://developer.fedoraproject.org/tech/languages/python/python-installation.html). Could you move it over there?

start/sw/web-app/django.md
```diff
@@ -10,130 +10,103 @@ Django is a high-level Python Web framework that encourages rapid development an
...
+$ cd venv
+$ source bin/activate
 ```
 
 When it's activated we will see indicator of same name as our virtualenv. 
 Now we will check if we really have empty environment:
 
 ```
-pip freeze
+$ pip freeze
```
pvalena: This is still pi usage, right?

start/sw/web-app/django.md
```diff
@@ -10,130 +10,103 @@ Django is a high-level Python Web framework that encourages rapid development an
...
 
-## How to install Django
+### How to install Django
 
 Now if you installed `virtualenv` do not leave it, we will install Django now. If you installed `virtualenv` use this command 
 inside your working directory with activated `virtualenv`:
 
 ```
-pip install django
+$ pip3 install django
```
pvalena: I think this should be on TOP of this page, as this is actually what user is looking for (not virtualenv), so maybe just mention this is to be done in already created virtualenv?

It should not be a complete guide, but rather a quickstart or reference.

start/sw/web-app/django.md
```diff
@@ -10,130 +10,103 @@ Django is a high-level Python Web framework that encourages rapid development an
...
 
 ```
-pip install django==1.10
-```
-
-For Python 3:
-
-```
-pip3 install django==1.10
+$ pip3 install django==1.10
```
pvalena: This is probably just a `pip` usage again.

start/sw/web-app/django.md
hroncok: If this ever gets mentioned without invoking virtualenv, make sure it is invoked with `--user` otherwise it will die on permission error and users tend to "solve" such problems with `sudo`.

start/sw/web-app/django.md
```diff
@@ -10,130 +10,103 @@ Django is a high-level Python Web framework that encourages rapid development an
...
-
 ```
-sudo dnf install python-django
-```
-
-or Python 3:
-
-```
-sudo dnf install python3-django
+$ sudo dnf install python3-django
```
pvalena: Could we prefer this instead of `pip`. What do we recommend user to do? Are there some (dis)advantages? I think this should be included.

_[ non-blocking ]_

start/sw/web-app/django.md
```diff
@@ -10,130 +10,103 @@ Django is a high-level Python Web framework that encourages rapid development an
...
 And you successfully started your first Django project.
 
-## Switching Virtualenvs
+### Switching Virtualenvs
 
 You can use `virtualenvwrapper` after installing it with usual command
+
 ```
-dnf install python-virtualenvwrapper
+$ sudo dnf install python3-virtualenvwrapper
```
pvalena: This is non django-related too, but rather general virtualenv usage, right?

hroncok: exactly, also a very specific way to do things. upstream recommends pipenv instead (we are working on packaging it into fedora and will update general python pages to mention pipenv)

pvalena:
I'd rather move all the content from the other django page in python subsection (deduplicate)  and leave just a link there (possibly [like this](https://developer.fedoraproject.org/tech/languages/php/php-frameworks.html)).

pvalena:
Since this is a fix of python page (not as much as updating of the content) I can accept the PR as is and create an issue instead.

Please let me know.
Thanks!

start/sw/web-app/django.md
pvalena: Yes, having enabled virtualenv is pretty much what the previous text says (L60). But, maybe it's worth mentioning you can use `--user` instead, if it's a reasonable alternative.

start/sw/web-app/django.md
pvalena: Ok, so maybe state that, instead of [L74](https://github.com/developer-portal/content/pull/242/files/f8fd7eb496b9aabf8a54fa6ed25f2e141ce612dc#diff-a0096ededfb8e6fc0c157dd79efbf8f8R74)?

start/sw/web-app/django.md
pvalena: Ok, wouldn't removing that line make sense then?

pvalena:
@thunderoy, any news on this?

pvalena:
Merging per my previous comment and created an issue for the rest.
Thanks!

PR #253, opened 2018-02-22
*csharp: Added vscode extension license.

RheaAyase:


RheaAyase:
Somehow it got created twice due to connection is my guess o_o

PR #256, opened 2018-03-02
use 'test-project' as the project name throughout

gavin-romig-koch:


pvalena:
Thanks!
I'm not sure what the author ment by `-2` there ... so lets assume it's a typo... would you mind updating also some of the commands from `F22` -> `F27`; `i386` -> `x86_64` etc.? You can just add a commit to his PR.
If not, just say so and I'll merge it.

pvalena:
(Why I'm asking is because you're probably following the guide, so you can test it on the go, which is really handful.)

PR #261, opened 2018-05-08
Remove Python 3.3

hroncok:
See https://lists.fedoraproject.org/archives/list/python-devel@lists.fedoraproject.org/thread/VTOQDUSGTMGOZOJHK4Y6GNVVLHA36QEC/


pvalena:
Thanks!

PR #271, opened 2018-07-13
Update rebase-helper docs

ferdnyc:
Did some cleanup, lots of formatting, and some minor copyediting in `rebase-helper.md` in the wiki:
* Indented code blocks that are placed under bullet headings
* Fenced code blocks with ` ```sh ` and ` ```python3 ` to enable (minimal) syntax highlighting
* Fenced more filenames / shell commands / package identifiers with backticks
* Removed backtick fencing from some headings and other non-code-literal words in favor of bolding or other (non-monospaced) decoration.
* Some minor grammatical cleanup (on the section headings, primarily).

### DIFF
--- a/deployment/maintain/rebase-helper.md
+++ b/deployment/maintain/rebase-helper.md
@@ -1,13 +1,10 @@
 ---
 title: Rebase helper
 subsection: maintain
-section: deployment
 order: 2
-description: >
-  Helps automate the process of updating a package to the latest upstream version
 ---
 
-# {{page.title}}
+# Rebase-helper
 Rebase-helper is a tool which helps package maintainers with updating package to the latest upstream version.
 It automates a lot of manual tasks the package maintainer usually does, when a new upstream version of a package is released.
 
--- a/deployment/maintain/rebase-helper.md
+++ b/deployment/maintain/rebase-helper.md
@@ -1,10 +1,13 @@
 ---
 title: Rebase helper
 subsection: maintain
+section: deployment
 order: 2
+description: >
+  Helps automate the process of updating a package to the latest upstream version
 ---
 
-# Rebase-helper
+# {{page.title}}
 Rebase-helper is a tool which helps package maintainers with updating package to the latest upstream version.
 It automates a lot of manual tasks the package maintainer usually does, when a new upstream version of a package is released.
 
@@ -45,44 +48,3 @@ If you do not want to be bothered, add the ``--non-interactive`` option to rebas
 
 After rebase-helper finishes, check the output.
 
-## Integrating rebase-helper with an upstream monitoring service
-
-This chapter demonstrates how to include rebase-helper in your upstream monitoring service. 
-If you would like to integrate rebase-helper into your upstream monitoring services, this chapter is for you.
-Rebase-helper provides an API which you can use either directly from Python, or directly from the command line.
-
-### Patch new upstream version and start scratch builds
-Example of patching new sources and starting scratch builds with fedpkg.
- This returns task_ids. The bash equivalents are included for comparison:
- 
-* **Python API**
-
-   ```python3
-   from rebasehelper.application import Application
-   cli = CLI(['--non-interactive', '--builds-nowait', '--buildtool', 'fedpkg', 'upstream_version'])
-   rh = Application(cli)
-   rh.set_upstream_monitoring() # Switch rebase-helper to upstream release monitoring mode.
-   rh.run()
-   ```
-
-* **BASH**
-
-   ```sh
-   $ rebase-helper --non-interactive --builds-nowait --buildtool fedpkg upstream_version
-   ```
-
-### Download logs and RPMs and compare with tools like ``abipkgdiff``
-
-* **Python API**
-
-   ```python3
-   cli = CLI(['--non-interactive', '--builds-nowait', '--buildtool', 'fedpkg', '--build-tasks', 'old_id,new_id'])
-   rh.run() # Downloads RPMs, logs and runs checkers and provides logs.
-   rh.get_rebasehelper_data() # Get all information about the results
-   ```
-
-* **BASH**
-
-   ```sh
-   $ rebase-helper --non-interactive --builds-nowait --buildtool fedpkg --build-tasks old_id,new_id
-   ```
--- a/deployment/maintain/rebase-helper.md
+++ b/deployment/maintain/rebase-helper.md
@@ -8,10 +8,10 @@ order: 2
 Rebase-helper is a tool which helps package maintainers with updating package to the latest upstream version.
 It automates a lot of manual tasks the package maintainer usually does, when a new upstream version of a package is released.
 
-## How to install rebase-helper
+## Installing rebase-helper
 Begin installation on Fedora using the ``dnf`` command:
 
-```
+```sh
 sudo dnf install -y rebase-helper
 ```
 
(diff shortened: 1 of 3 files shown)


deployment/maintain/rebase-helper.md
```diff
@@ -8,10 +8,10 @@ order: 2
 Rebase-helper is a tool which helps package maintainers with updating package to the latest upstream version.
 It automates a lot of manual tasks the package maintainer usually does, when a new upstream version of a package is released.
 
-## How to install rebase-helper
+## Installing rebase-helper
```
pvalena: I'd preffer `Install` instead of `Installing`.

deployment/maintain/rebase-helper.md
```diff
@@ -20,29 +20,32 @@ These programs are installed automatically as dependencies of rebase-helper.
...
 
-## How does the rebase-helper work?
+## How does rebase-helper work?
 Rebase-helper workflow can be summarized in following steps:
 
 - Downloads an archive with new sources.
 - Rebases downstream patches on top of the latest upstream version using git rebase command.
-- If patching was successful, it builds two sets of RPMs. One using the ``old`` sources and a second one using the ``new`` sources. Available build tools are ``mock``, ``rpmbuild``, ``fedpkg``.
-  - If ``new`` RPMs build fails, downloads logs like build.log and root.log
+- If patching was successful, it builds two sets of RPMs. One using the **``old``** sources and a second one using the **``new``** sources. Available build tools are ``mock``, ``rpmbuild``, ``fedpkg``.
```
pvalena: Regarding the markdown changes, I wouldn't put both ``
and ** there.  Choose one, please.

deployment/maintain/rebase-helper.md
```diff
@@ -52,34 +55,34 @@ Rebase-helper provides an API which you can use either directly from Python, or
...
-rh.set_upstream_monitoring() # Switch rebase-helper to upstream release monitoring mode.
-rh.run()
-```
+   ```python3
+   from rebasehelper.application import Application
+   cli = CLI(['--non-interactive', '--builds-nowait', '--buildtool', 'fedpkg', 'upstream_version'])
+   rh = Application(cli)
+   rh.set_upstream_monitoring() # Switch rebase-helper to upstream release monitoring mode.
+   rh.run()
+   ```
```
pvalena: This has unnecessary indent, right?

deployment/maintain/rebase-helper.md
```diff
@@ -52,34 +55,34 @@ Rebase-helper provides an API which you can use either directly from Python, or
...
-```
+   ```sh
+   rebase-helper --non-interactive --builds-nowait --buildtool fedpkg upstream_version
+   ```
 
-### Download logs and RPMs and compare with tools like abipkgdiff
+### Download logs and RPMs and compare with tools like ``abipkgdiff``
 
-* ``Python API``
+* **Python API**
```
pvalena: It looks like there's a lot of directions in one page. Would you mind splitting it into multiple pages?

deployment/maintain/rebase-helper.md
```diff
@@ -52,34 +55,34 @@ Rebase-helper provides an API which you can use either directly from Python, or
...
-rh.get_rebasehelper_data() # Get all information about the results
-```
+   ```python3
+   cli = CLI(['--non-interactive', '--builds-nowait', '--buildtool', 'fedpkg', '--build-tasks', 'old_id,new_id'])
+   rh.run() # Downloads RPMs, logs and runs checkers and provides logs.
+   rh.get_rebasehelper_data() # Get all information about the results
+   ```
 
-* ``BASH``
+* **BASH**
```
pvalena: Wouldn't `command line` be more accurate here?

deployment/maintain/rebase-helper.md
```diff
@@ -8,10 +8,10 @@ order: 2
 Rebase-helper is a tool which helps package maintainers with updating package to the latest upstream version.
 It automates a lot of manual tasks the package maintainer usually does, when a new upstream version of a package is released.
 
-## How to install rebase-helper
+## Installing rebase-helper
 Begin installation on Fedora using the ``dnf`` command:
 
-```
+```sh
 sudo dnf install -y rebase-helper
```
pvalena: Additionally, would you mind removing the `-y` argument here?

deployment/maintain/rebase-helper.md
```diff
@@ -20,29 +20,32 @@ These programs are installed automatically as dependencies of rebase-helper.
...
+Let's say we want to rebase a package foobar from ``foobar-1.2.0`` to ``foobar-1.2.1`` using rebase-helper:
 
-To update to the latest upstream version execute command:
+```sh
+# Change to the location of foobar.spec and other package components (cloned dist-git dir), e.g.
+cd $HOME/rpmbuild/REPOS/foobar
 
-```
+# Update to the selected upstream version
 rebase-helper 1.2.1
```
pvalena: What is this actually for?

pvalena: Actually, would you mind prefixing these commands with `$`?

deployment/maintain/rebase-helper.md
ferdnyc: Fair enough.

deployment/maintain/rebase-helper.md
ferdnyc: Agreed, those always made me uncomfortable.

deployment/maintain/rebase-helper.md
pvalena: Well, AFAICT it's because of the syntax hilighting of key words, see resulting page:

https://developer.fedoraproject.org/deployment/maintain/rebase-helper.html

Thinking about it some more, maybe we could live the backtics instead

deployment/maintain/rebase-helper.md
ferdnyc: I played with that at one point, and determined that the syntax highlighting _really_ didn't handle that in a way that felt very readable. And it has the cost of making it a lot harder to cut-and-paste the commands, if someone wants to use them directly out of the document. I'll make the change, but IMHO of these two formats:

```sh
# Change to the location of foobar.spec and other package components (cloned dist-git dir), e.g.
$ cd $HOME/rpmbuild/REPOS/foobar

# Update to the selected upstream version
$ rebase-helper 1.2.1
```

```sh
# Change to the location of foobar.spec and other package components (cloned dist-git dir), e.g.
cd $HOME/rpmbuild/REPOS/foobar

# Update to the selected upstream version
rebase-helper 1.2.1
```
...The second is actually quite a bit clearer and more useful. (Again, purely IMHO.)

There's also `console` syntax highlighting, which is meant for console typescripts... but it doesn't style comments at all, and still fails to style the shell prompt usefully. Really it's more geared towards showing _output_ along with input, which we aren't doing.

```console
# Change to the location of foobar.spec and other package components (cloned dist-git dir), e.g.
$ cd $HOME/rpmbuild/REPOS/foobar

# Update to the selected upstream version
$ rebase-helper 1.2.1
```
...So, no real wins there, but a few losses.

deployment/maintain/rebase-helper.md
ferdnyc: Oh, not at all, IMHO. "**Python API**" and "**BASH**" are two items in a bulleted list, the indent places the examples at the same level (inside their corresponding bullet points).

deployment/maintain/rebase-helper.md
pvalena: If we're to enhance the readability we'll do that site-wide. Feel free to submit an ehnancement (preferrably CSS or similar). As of now, we prefix all `commands` with `$` (see our [Contributing Guidelines](http://developer-portal.github.io/contributing/write-general-writing-hints.html)), so please do include it.

Also, FYI: the jekyl-built website looks different compared to Githubs-Markdown display.

deployment/maintain/rebase-helper.md
ferdnyc: We've kind of moved beyond "directions" by this point in the document, TBH. 

Everything from "Integrating rebase-helper with an upstream monitoring service" down is a separate topic. I think it's fine to be its own separate page, but it's also not really "directions". It's more a rough sketch of possible pseudocode that might be employed for that purpose. 

You asked elsewhere if the code has been tested on current Fedora. As far as I know, this code _never_ worked on **any** version of Fedora / fedpkg. I had nothing to do with drafting the original document, but my impression is that whoever wrote it just jotted down some half-formed thoughts and never really fleshed them out. One of my original commits included this note:

> The syntax highlighting was also very useful as it highlighted several typos in the API examples. This fixes some more of them, as best I could _guess_ what they were supposed to be. But there are still some issues, since (for example) `fedpkg` isn't a legal argument to `--buildtool`. (The F28 `rebase-helper` says: "`ERROR: argument --buildtool: invalid choice: 'fedpkg' (choose from 'copr', 'koji', 'mock', 'rpmbuild')`".) But I only corrected what I could be reasonably sure of.

I would be very surprised if any of the code in these examples was 100% usable as-is. But, while I can correct the formatting, I can't correct the code. I didn't write it.

I'll part out the "Integrating..." bit into its own page, though.

deployment/maintain/rebase-helper.md
ferdnyc: I don't... _think_ so. My impression is that the examples there were meant for `/bin/sh` shell scripting, when integrating into a monitoring service controlled by shell scripts. They weren't envisioned as interactive commands. But, that's purely my guess of the original author's intent, based on the context.

ferdnyc: (In fact, while I added `$` prompts to the start of these commands, too, I'm now thinking that may have been a mistake, for exactly those reasons.)

deployment/maintain/rebase-helper.md
pvalena: Well, even in case that is reflected on the resulting page properly, I'd prefer using (sub-)sub-headlines instead of those bullet points.

deployment/maintain/rebase-helper.md
pvalena: Thanks.
So fixing the code should be on issues list.
Every code snippet on this web should work (at least at some point), otherwise it's useless.

pvalena:
``This branch has no conflicts with the base branch when rebasing``

deployment/maintain/rebase-helper.md
ferdnyc: That's the odd thing, though. `old` and `new` _aren't_ actually keywords, as far as I can tell. They're closer to argument placeholders, in fact I'm wondering if they shouldn't be italicized instead.

deployment/maintain/rebase-helper.md
ferdnyc: Mmm, actually in moving that part to a separate page I ended up raising everything a heading level, so now they will just be sub-sub. (Previously, the titles of the sectinons they occupy were already sub-sub. Now they're just subheads.) ...Done.

pvalena:
I think you're right. Everything belonging to `Integrating with upstream monitoring service` is meant as a script snippet. 
It's just an example though, and don't think anyone ever used it. Feel free to change it any way you wish.

ferdnyc:
I think I've now addressed everything we discussed, let me know if any more changes are needed. Thanks for taking the time to go through it!

pvalena:
Looks great, thanks!

PR #277, opened 2018-08-21
Update and extend section about Rust

msehnout:
Changes:
 * Main page reflects how to get started rather than how to install the compiler
 * Cargo now defaults to binary projects instead of libraries (`cargo new`)
 * Rust SIG
 * useful links

### DIFF
--- a/tech/languages/rust/further-reading.md
+++ b/tech/languages/rust/further-reading.md
@@ -1,0 +1,42 @@
+---
+title: Further reading
+subsection: rust
+order: 4
+---
+
+# Learning resources
+
+There is an official book describing the Rust programming language in detail.
+https://doc.rust-lang.org/book/2018-edition/index.html
+
+More examples of the concepts discussed in the book are available in this collection:
+https://doc.rust-lang.org/rust-by-example/
+
+When you finish with the language itself, there is an ongoing effort to create a set of examples of using core libraries for concurrency, data serialization, network programming, etc.
+https://rust-lang-nursery.github.io/rust-cookbook/intro.html
+
+Finally, you can find community pages with a list of resources such as this one:
+https://github.com/ctjhoa/rust-learning
+
+
+# API documentation
+
+Standard library documentation:
+https://doc.rust-lang.org/std/
+
+Documentation for crates available at `crates.io`:
+https://docs.rs/
+
+# IDE recommendations
+
+There are plugins for popular editor such as Vim, Atom or Gnome Software which are based on the Rust Language Server (RLS) implementation. Alternatively, IntelliJ IDEA provides their own, custom plugin. Unfortunatelly, the debugging support is available only in paid CLion IDE.
+https://areweideyet.com/
+
+# Other useful links
+
+If you want to start with some web applications, a list of useful frameworks is available here:
+http://www.arewewebyet.org/
+
+If you are brave enough and you want to try the nightly compiler edition (or you just want to try the awesome [Rocket framework](https://rocket.rs/)), it is recommended to use rustup:
+https://rustup.rs/
+
--- a/tech/languages/rust/rust-sig.md
+++ b/tech/languages/rust/rust-sig.md
@@ -1,0 +1,10 @@
+---
+title: Rust SIG
+subsection: rust
+order: 5
+---
+
+# Rust SIG
+
+Rust Special Interest Group takes care of anything related to Rust in Fedora distribution. You can find more information at the official web pages:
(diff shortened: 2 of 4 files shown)


tech/languages/rust/further-reading.md
```diff
@@ -0,0 +1,42 @@
+---
+title: Further reading
+subsection: rust
+order: 4
+---
+
+# Learning resources
```
pvalena: I wouldn't start with this.
It can be at the bottom of some page as `references` (I'm not sure about the book though).
This 'portal' aims to have `quickstarts` and `references` useful for *Developers on Fedora* (Fedora-specifics being the main part). Please give a quick check to these:

https://fedoraproject.org/wiki/Websites/Developer#About_The_Project
http://developer-portal.github.io/contributing/write-introduction.html

tech/languages/rust/further-reading.md
```diff
@@ -0,0 +1,42 @@
...
+https://github.com/ctjhoa/rust-learning
+
+
+# API documentation
+
+Standard library documentation:
+https://doc.rust-lang.org/std/
+
+Documentation for crates available at `crates.io`:
+https://docs.rs/
```
pvalena: Missing link. Also a `reference`.

tech/languages/rust/further-reading.md
```diff
@@ -0,0 +1,42 @@
...
+
+# IDE recommendations
+
+There are plugins for popular editor such as Vim, Atom or Gnome Software which are based on the Rust Language Server (RLS) implementation. Alternatively, IntelliJ IDEA provides their own, custom plugin. Unfortunatelly, the debugging support is available only in paid CLion IDE.
+https://areweideyet.com/
```
pvalena: It'd be best if you can make some recommendation / guide from this. Ask yourself a question "how can user do this on Fedora?"

tech/languages/rust/rust-sig.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Rust SIG
+subsection: rust
+order: 5
+---
+
+# Rust SIG
+
+Rust Special Interest Group takes care of anything related to Rust in Fedora distribution. You can find more information at the official web pages:
+https://docs.pagure.org/fedora-rust.sig/index.html
```
pvalena: Great!
But please make this a Markdown link (with a text, instead of just an url).

tech/languages/rust/further-reading.md
msehnout: This is a whole section with references. I think it is much more useful to include links to the upstream project than to write something new. This helps by pointing developers in the right direction.

tech/languages/rust/further-reading.md
msehnout: (s)he can install atom/vscode/intellij ... none of them is packaged for Fedora. Maybe it would be more useful to write a section about the installation of these editors/ides. The only editor available in Fedora is Vim (maybe Emacs, if anyone is really using it), which can use the language server implementation with some plugins, but I don't want to write a tutorial on that, because it will break with the next upstream release of vim, vundle, rls or whatever.

tech/languages/rust/further-reading.md
pvalena: Yes, I can see that. I'd would prefer something like this, instead:

 - "main page" (basic Fedora install + basic info + general references)
 - "arbitrary-topic-page" (could be "cargo" or "ides", basic install & info, topic-related references  -- one topic per page!)
 - "sig-or other page" (contacts)

tech/languages/rust/further-reading.md
pvalena: Well, that's important information and could be standalone section indeed. 
No extensive "tutorial" is needed (or wanted) here. What we need is just basic info to send a developer in the right direction (for him to find and get the right "thing" he/she needs and some sources/references for further inqueries).

tech/languages/rust/further-reading.md
pvalena: Note: we also prefer to link the content instead of duplicating it. It's not about the content but a structure of it.

pvalena: IOW everyone will skip this page if it's not short and per-purpose structured.

pvalena: This could be an "About page". With some short intro, links to the other sections (incl. the "install rust" one), and references to books etc. at the bottom.

pvalena:
Thank you for considering the changes.

PR #279, opened 2018-08-27
There will no more virtualenv-X

hroncok:
Updated to support both F27/28 and F29/30

Once 28 is EOL, we can chnage the command to `dnf install virtualenv`

See https://src.fedoraproject.org/rpms/python-virtualenv/pull-request/5


pvalena:
Thanks for the PR!

pvalena:
_Note: FDP aims to be valid for the latest stable Fedora (mainly)._

Issue #283, opened 2018-08-28
Multiple Python interpreters: Mention Python 2.6 is not supported with tox


hroncok:
Since tox 3, Python 2.6 is not supported. Instead of patching that back in like we did with virtualenv, we should just mention here that this works for 2.7+ only.

https://developer.fedoraproject.org/tech/languages/python/multiple-pythons.html

Section "Getting it and running it all with tox".

I'll change the text, but not now, this is a note for me not to forget.

hroncok:
(Assign me please.)

PR #284, opened 2018-08-31
Add a note about tox 3 not supporting Python 2.6

hroncok:
Fixes https://github.com/developer-portal/content/issues/283

This is only relevant for Fedora 29+, but it is written in a way that it can be merged even now.


pvalena:
LGTM, thanks!

PR #288, opened 2018-09-15
about.md

luiscontrerasdo:
Now update for version 4 of MongoDB, even MongoDB is in version 4.2 I believe having minimum 4.0 installed is a good idea, and avoiding older version like 3.x


pvalena:
Thanks!
LGTM.

PR #289, opened 2018-09-16
Add note about vagrant-libvirt qemu performance, update upstream URL

jwflory:
These are minor edits to the Vagrant libvirt page. I added a note about the performance hits (because this happened to me on my Fedora 28 workstation) and I also updated a project's GitHub URL since the upstream moved.


pvalena:
Thanks!

PR #309, opened 2019-03-12
Mongotron warning

sergushakov:
project really dead and buggy. may be someone know better monogo-gui, and may be we must delete mongotron section at all


pvalena:
You're right, let's use https://github.com/hatamiarash7/MongoDB_Admin/wiki/1.-Getting-Start instead. Have you tried it?

PR #312, opened 2019-04-07
Updated gtk.md

Macr0Nerd:
I just updated the command for the necessary installation. `webkitgtk3-devel` was retired and replaced with `webkit2gtk3-devel`. I just updated the command with that. Also, it is listed under gtk-3.0, so I mentioned that.


pvalena:
LGTM, install tested on Rawhide.

Thanks!

PR #313, opened 2019-04-11
Instructions to install local Web Server in CONTRIBUTING.md

cbillowes:
I want to contribute but don't want to deliver poorly formatted markdown files. I needed an http server which took some time to find. I wanted a lightweight server which did not need configuration in the working directory. I thought it would be useful to share this with anyone who would like to contribute.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,2 +1,9 @@
 # Contribution to Fedora Developer Portal
 Information about contributing is at [developer-portal.github.io](https://developer-portal.github.io/).
+
+### Local http server
+You could use [nmp](https://www.npmjs.com/) to download and install a lightweight http server to view your changes.
+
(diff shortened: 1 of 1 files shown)


pvalena:
Hello!
This is use case is actually unintended, as you need to run Jekyll specifically to build the site and it can serve the files as well.

Please help me to make this clear in the contribution guide.

pvalena:
IOW we do not wan't to use anything else other than Jekyll to render the pages into html.
_ _ _ _
On this page of our / Contributing guide /
https://developer-portal.github.io/contributing/write-edit-and-preview.html
there is:
```
Testing your changes locally

You can build and run the website on your machine to ...
```

Which points the reader to / Developement guide /
 https://github.com/developer-portal/website/blob/master/DEVELOPMENT.md

Which actually has the instructions to run the local instance (or in Docker or Vagrant). It should be up-to-date, so please let me know if it does not work for you; or how we can make it more straightforward (PRs are very welcome).

PR #315, opened 2019-06-02
Update vagrant-libvirt docs

lentzi90:
The QEMU connection used was changed in the vagrant-libvirt package. This updates the documentation to reflect that change.

### DIFF
--- a/tools/vagrant/vagrant-libvirt.md
+++ b/tools/vagrant/vagrant-libvirt.md
@@ -61,7 +61,11 @@ end
 ```
 
 Note that this will require the administrator password to run your vagrant boxes.
-For more information on this, see the [upstream documentation](https://github.com/vagrant-libvirt/vagrant-libvirt#qemu-session-support)
+For more information on this, see the [upstream documentation](https://github.com/vagrant-libvirt/vagrant-libvirt#qemu-session-support).
+The QEMU session will also affect where the vagrant boxes are stored (`~/.local/share/libvirt/images/` or `/var/lib/libvirt/images/`).
+This means that if you already have initialized boxes and change the QEMU connection, **things will break**.
+
+More information on the difference between the system and session connections can be found [here](https://wiki.libvirt.org/page/FAQ#What_is_the_difference_between_qemu:.2F.2F.2Fsystem_and_qemu:.2F.2F.2Fsession.3F_Which_one_should_I_use.3F).
 
 ## Using QEMU system connection without password prompts
 
--- a/tools/vagrant/vagrant-libvirt.md
+++ b/tools/vagrant/vagrant-libvirt.md
@@ -41,10 +41,31 @@ end
 
 Read more on the `vagrant-libvirt` configuration in the [upstream documentation](https://github.com/vagrant-libvirt/vagrant-libvirt).
 
-## Using libvirt from Vagrant without password prompts
+## QEMU session or system connection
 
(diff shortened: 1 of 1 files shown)


pvalena:
Thanks for writing this!

Could you possibly mention the `Vagrantfile in your Vagrant home directory (defaults to ~/.vagrant.d). This lets you specify some defaults for your system user. ` Vagrantfile? This change would propagate to all of the users' Vagrantfiles.

Also, this change (in one direction or the other) will render all of the already initialized boxes unavailable (even the base ones); in either `~/.local/share/libvirt/images/` or `/var/lib/libvirt/images/`.

lentzi90:
Oh I didn't know about that, thanks for pointing it out! I can certainly add it, but I'm wondering if it should go in a separate pull request and perhaps in a separate file as it is not specific to the libvirt provider?

pvalena:
Good thinking.
I agree with creating `Vagrantfile` page, giving (simple) examples etc. and let's just link to it(can you do that, or should I?).

The info about images location (_and QEMU Session_) mentioned in my comment is `vagrant-libvirt`-specific, so I'd include it into this page. Could you please add it?

lentzi90:
I think I can do a `Vagrantfile` page next week, but I'm not sure about the libvirt images. What did you have in mind? Should it just be a note warning the user about changes to the "default" Vagrantfile?

pvalena:
> Should it just be a note warning the user about changes to the "default" Vagrantfile?

Yes, a short note, and please refer to the upstream documentation.

Thanks!

pvalena:
Actually: not about the default Vagrantfile, but about the QEMU Session change, as it can affect where the images / base images / running images are stored or loaded from (on the libvirt side) and may potentially disrupt user's environment.

lentzi90:
I added a bit more now. Is there a place with more documentation on the libvirt specifics? I had a hard time finding documentation to link to. For Fedora specifics I found [this](https://docs.fedoraproject.org/en-US/quick-docs/getting-started-with-virtualization/), but it doesn't really go into the session connection and storage differences. Then there is the [libvirt docs](https://libvirt.org/docs.html) but I found it hard to navigate and perhaps a bit too deep?

pvalena:
> I added a bit more now. Is there a place with more documentation on the libvirt specifics? I had a hard time finding documentation to link to. For Fedora specifics I found [this](https://docs.fedoraproject.org/en-US/quick-docs/getting-started-with-virtualization/), but it doesn't really go into the session connection and storage differences. Then there is the [libvirt docs](https://libvirt.org/docs.html) but I found it hard to navigate and perhaps a bit too deep?

Well, maybe the vagrant-libvirt project then?
https://github.com/vagrant-libvirt/vagrant-libvirt/

Complexity of libvirt is why vagrant exists in the first place. I've writen an email about this on Fedora-devel, co you can cross-check for the information:
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/AQX4CTGHBU4QQZIQ7HU2EVFGSGWFDZAN/

Thank you for your PR.

pvalena:
Please give it a look, so we can merge it (it does not have to be perfect, just give it a byte of love).

_ _ _ _ 
_Note: The release will be soon (~1-2weeks), and after that there will be maintenance (decomissioning) of the release infrastucture(no updates for some time). The website updates still need some manual handling, but openshift will change that._

PR #320, opened 2019-07-30
Updated position of --scratch

tachoknight:
Putting `--scratch` before `build` results in the error `koji: error: no such option: --scratch`; putting it _after_ `build` works.


pvalena:
You're right, thanks!

PR #327, opened 2020-02-12
Enable nfs3 service in firewall

bocekm:
`nfs ` service means NFSv4 in firewalld. `nfs3` is for NFSv3. I don't know in which version of Fedora the nfs3 firewalld service has been introduced. But it was there since at least F29.

NFSv3 is used by vagrant as default at this moment (https://www.vagrantup.com/docs/synced-folders/nfs.html):
> nfs_version (string | integer) - The NFS protocol version to use when mounting the folder on the guest. This defaults to 3.

### DIFF
--- a/tools/vagrant/vagrant-nfs.md
+++ b/tools/vagrant/vagrant-nfs.md
@@ -19,7 +19,8 @@ $ sudo dnf install nfs-utils && sudo systemctl enable nfs-server
 Afterwards enable `nfs`, `rpc-bind` and `mountd` services for `firewalld`:
 
 ```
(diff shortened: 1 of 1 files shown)


bocekm:
When will the merged PR translate into the update of https://developer.fedoraproject.org/tools/vagrant/vagrant-nfs.html?

pvalena:
_Note: The release will be soon (~1-2weeks), and after that there will be maintenance (decomissioning) of the release infrastucture(no updates for some time). The website updates still need some manual handling, but openshift will change that._

PR #328, opened 2020-02-14
Use libvirt firewall zone for nfs in vagrant

bocekm:
- the libvirt provider interface virbr0 is using the `libvirt` firewall zone
  https://libvirt.org/firewall.html#fw-firewalld-and-virtual-network-driver

### DIFF
--- a/tools/vagrant/vagrant-nfs.md
+++ b/tools/vagrant/vagrant-nfs.md
@@ -19,10 +19,10 @@ $ sudo dnf install nfs-utils && sudo systemctl enable nfs-server
 Afterwards enable `nfs`, `rpc-bind` and `mountd` services for `firewalld`:
 
 ```
-$ sudo firewall-cmd --permanent --add-service=nfs3 \
-    && sudo firewall-cmd --permanent --add-service=nfs \
-    && sudo firewall-cmd --permanent --add-service=rpc-bind \
-    && sudo firewall-cmd --permanent --add-service=mountd \
+$ sudo firewall-cmd --permanent --zone=libvirt --add-service=nfs3 \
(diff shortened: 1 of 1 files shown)


bocekm:
This needs more investigation:
- Why wasn't mentioning the zone needed on Fedora 30? It's necessary since F31.
- What's the default provider on Fedora? This update is libvirt-specific.

Related: https://lukas.zapletalovi.com/2020/01/setting-a-firewalld-zone-for-libvirt-network.html

pvalena:
Wow, thanks for investigation!

This is probably some hardening on libvirt side, which should be handled on 'vagrant-libvirt' side.

On Fedora is the default provider (also shipped with Fedora, thus tested).

pvalena:
(I'll bring up the issue on vagrant-libvirt.)

jackorp:
> (I'll bring up the issue on vagrant-libvirt.)

Is there any update concerning this? Otherwise, I think we can merge this.

pvalena:
I haven't found anything that would concern `vagrant-libvirt` provider, specifically.

From Fedora POV, I've tested it (but I already had it enabled previsouly) and I'm unable to verify it works, as NFS breaks for me with this Vagrantfile:

```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "fedora/34-cloud-base"
  config.vm.synced_folder ".", "/vagrant", type: "nfs"

  config.vm.provider :libvirt do |libvirt|
    libvirt.nested = false
    libvirt.memory = 2048
  end
end
```
Could you verify? If it works for you, feel free to merge it.

jackorp:
Verified, it does work. I'll merge it.
Thanks!

PR #329, opened 2020-02-18
gradle not part of fc31 repo(s)

shoeper:


pvalena:
Ah then please remove it completely then, or at least mark it clearly (heading) deprecated.

We aim to provide up-to-date info for developers which is latest stable Fedora.

shoeper:
Ok, I currently don't know what the recommended way to install is. 

https://bugzilla.redhat.com/show_bug.cgi?id=1657836

PR #332, opened 2020-03-03
multiple pythons: add CPython 3.9

vstinner:


vstinner:
cc @Dormouse759 @hroncok

PR #337, opened 2020-05-06
Updating the link to the official guide for .NET Core.

phillipsj:
The getting started link to the official guide is using an old link that results in a 404. I have updated it to the new location.


phillipsj:
This fixes #338

pvalena:
Thanks!

PR #350, opened 2020-10-02
Added content on writing unit tests and mocks in C with AUM

c8y3:
I have written a succinct introduction to unit tests in C with AUM (https://github.com/airbus-cyber/aum). 
AUM is a C unit testing library which focuses on the easy definition of mocks.
I explain the reason for doing unit tests, the structure of a test, the notion of mocks and list some best practices. All the examples are written using the AUM API.

### DIFF
--- a/tech/languages/c/unit_tests_and_mocks.md
+++ b/tech/languages/c/unit_tests_and_mocks.md
@@ -1,0 +1,60 @@
+---
+title: Unit Tests and Mocks
+subsection: c
+order: 5
+---
+
+# Unit tests
+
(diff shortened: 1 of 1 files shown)


pvalena:
Hello. It looks good!

However, I'd like it to look more Fedora-friendly, as this is aimed solely at Fedora users.

E.g. https://github.com/developer-portal/content/blob/master/tech/languages/c/gtk.md

 - Starts with overview (we really aim at short quickstarts)
 - Continues with `dnf install` (fedora-specific) .. could be replaced by COPR. f.e..
 - Example (simple usage)
 - References documentation. 

Also, this could be possibly generalized- are there any other mocking libraries (just to list them for comparison)?

PR #357, opened 2020-12-10
Set npm global modules path

damiantoczek:
npm will throw an error out of the box when installing global modules like _nodemon_ etc.
I've added a solution how to set the new npm global module directory.


pvalena:
Could you please add the comment (or why it's needed) into the doc as well?

pvalena:
Thank you for your PR!

pvalena:
LGTM, thanks!

PR #365, opened 2021-03-07
Update mongodb/about.md

muayyad-alsadi:
update version and use el8 repo instead of amazon linux (because it has better systemd support)


pvalena:
Thanks!

PR #366, opened 2021-03-24
Replace openjdk15 on java-installation.md

MohammaJavadTD:
Replace openjdk15 instead of openjdk11 in the text


pvalena:
LGTM. Thanks!

PR #368, opened 2021-04-06
With SELinux, you have to append `:z` to volumes

vonbrand:


tools/docker/compose.md
```diff
@@ -79,13 +79,12 @@ web:
...
    - "8000:8000"
   volumes:
-   - ./awesome_web:/code
   links:
    - db
 db:
   image: fedora-postgresql
   volumes:
-   - ./db:/var/lib/pgsql/data
+   - ./db:/var/lib/pgsql/data:z
```
pvalena, line 87: If I understand it correctly, this is for sharing between containers, right? _(`:z` -> correct)_

FYI: for sharing outside of container, you need capital `:Z`.

tools/docker/compose.md
```diff
@@ -79,13 +79,12 @@ web:
   ports:
    - "8000:8000"
   volumes:
-   - ./awesome_web:/code
```
pvalena, line 82: You've removed the volume share altogether. Can you elaborate?

pvalena (changes requested):
Thanks for the PR!
Please see the respective comments.

renich:
It should be noted that :z and :Z have different behaviors:

```
Labeling Volume Mounts

Labeling systems like SELinux require that proper labels are placed on volume content mounted into a container. Without a label, the security system might prevent the processes running inside the container from using the content. By default, Podman does not change the labels set by the OS.

To change a label in the container context, you can add either of two suffixes :z or :Z to the volume mount. These suffixes tell Podman to relabel file objects on the shared volumes. The z option tells Podman that two containers share the volume content. As a result, Podman labels the content with a shared content label. Shared volume labels allow all containers to read/write content. The Z option tells Podman to label the content with a private unshared label.

Note: Do not relabel system files and directories. Relabeling system content might cause other confined services on your machine to fail. For these types of containers we recommend disabling SELinux separation. The option --security-opt label=disable disables SELinux separation for the container. For example if a user wanted to volume mount their entire home directory into a container, they need to disable SELinux separation.
```

https://docs.podman.io/en/latest/markdown/podman-run.1.html

PR #369, opened 2021-04-06
The container is called `awesome_web_db_1`

vonbrand:


pvalena:
Thanks for fixing the typo!

PR #371, opened 2021-04-13
docker: Drop information about switching back to cgroup v1

Conan-Kudo:
Docker fully supports cgroup v2 with Docker version 20.10.


pvalena:
Thanks!
Is this valid for all stable Fedoras (F33+) ?

Conan-Kudo:
@pvalena Yes, since you're telling people to get Docker from Docker's repo rather than using `moby-engine` in Fedora.

pvalena:
Oh, right. Thanks.

PR #376, opened 2021-04-14
Update Python guide

encukou:
- Fully switch to Python 3
- Mark instructions for opening the Terminal as specific to GNOME

### DIFF
--- a/tech/languages/python/python-installation.md
+++ b/tech/languages/python/python-installation.md
@@ -14,8 +14,8 @@ Python 3 is already pre-installed on Fedora. Let's use it!
 
 ## Running Python
 
-1. Open the _Terminal_ (press `Alt` + `F1`, type _Terminal_ and click on the black square icon or just press `Enter`).
-2. To run Python 3 type `python3`. You should see something like this:
+1. Open your _terminal_. (In GNOME, press `Alt` + `F1`, type _Terminal_ and press <kbd>Enter</kbd>.)
+2. To run Python, type `python` <kbd>Enter</kbd>. You should see something like this:
 
 ```python
 Python 3.9.0 (default, Oct  6 2020, 00:00:00) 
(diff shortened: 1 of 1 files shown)


hroncok:
Do we assume `/usr/bin/python` is installed?

encukou:
I think we can assume that.

pvalena:
@hroncok do you want commit rights to `developer-portal/content` :) ?

I think you can merge it yourself if you find it ok.

tech/languages/python/python-installation.md
```diff
@@ -14,8 +14,8 @@ Python 3 is already pre-installed on Fedora. Let's use it!
 
 ## Running Python
 
-1. Open the _Terminal_ (press `Alt` + `F1`, type _Terminal_ and click on the black square icon or just press `Enter`).
-2. To run Python 3 type `python3`. You should see something like this:
+1. Open your _terminal_. (In GNOME, press `Alt` + `F1`, type _Terminal_ and press <kbd>Enter</kbd>.)
```
pvalena, line 17: If we're to guide <developers> how to run a terminal, it should be a standalone guide, for various desktop spins. I don't find it fitting for Python-installation / running python guide.

I'd like to assume everyone knows how to run terminal and a command in it. Otherwise half of the developer portal would need a makeover :) .

hroncok: This has been hare already, the change does not make it nay worse.

pvalena: Doing nothing is not better than doing something to make it better ;). Or do you want me to simply remove it (I can open a PR)?

hroncok: I like the note as proposed. We can share the page with GNOME/Fedora/Linux beginners and whoever knows how to opena  terminal simply skips it.

encukou: > If we're to guide how to run a terminal, it should be a standalone guide, for various desktop spins.

Sounds good; I'll happily link to it instead, once it's up!

But before then, as Miro says, I'd like to make Python developers coming from another OS feel welcome :)

pvalena: > I like the note as proposed. We can share the page with GNOME/Fedora/Linux beginners and whoever knows how to opena terminal simply skips it.

Linux beginner + Python beginner does not look like our target audience to me :), but I may be wrong. Python is a first language for many developers, so feel free to keep it if you like it!

But I definitely would not make it a general rule/guidance for Developer Portal.

tech/languages/python/python-installation.md
```diff
@@ -91,11 +91,5 @@ As an alternative, you can create the environment with `--system-site-packages`
 ### What next?
 
 * [Python homepage](https://www.python.org/)
-* [Python 3 Documentation](https://docs.python.org/3/)
+* [Python Documentation](https://docs.python.org/3/)
```
hroncok, line 94: What about this?

```suggestion
* [Python Documentation](https://docs.python.org/)
```

hroncok (approved):

PR #379, opened 2021-04-28
Update instructions for pip install

encukou:
- Add pip installation. The Python 3.10 package will no longer recommend `pip`, so the tool might not always be installed system-wide. (It will still be added to new virtualenvs, however.)
- Add a section on updating packages.

### DIFF
--- a/tech/languages/python/pypi-installation.md
+++ b/tech/languages/python/pypi-installation.md
@@ -56,15 +56,30 @@ If you no longer need it, deactivate it and delete it with `rm -rv project_venv`
 
 Sometimes, you don't want to use a virtual environment.
 For example, you're installing a tool you will use in multiple projects, but the tool is not available in Fedora repositories.
-In such cases, install with the `--user` option.
+In such cases, you might need to install `pip` if you do not already have it system-wide:
+
+```bash
+$ sudo dnf install python3-pip
+```
+
+Then, use pip's `--user` option.
 For example, to install a Python implementation of `cowsay` (rather than the original Perl one available in Fedora), run:
 
 ```bash
 $ python -m pip install --user cowsay
 ```
 
-If you're feeling lucky or impatient, you can use the shorter command `pip install cowsay`, without the `python -m`.
+If you're feeling lucky or impatient, you can use the shorter command `pip install --user cowsay`, without the `python -m`.
 Usually, this will install using the correct version of Python.
+
+## Updating packages
+
+Since software installed using `pip` are not part of Fedora, it is your responsibility to keep it updated as needed.
+However you ran `pip install`, running it again with `--update` will update to the latest release. For example:
+
+```bash
+(project_venv) $ python -m pip install --update requests
+```
 
 
 ### What next?
--- a/tech/languages/python/pypi-installation.md
+++ b/tech/languages/python/pypi-installation.md
@@ -76,7 +76,7 @@ Usually, this will install using the correct version of Python.
 
 Since software installed using `pip` is not part of Fedora, it is your responsibility to keep it updated as needed.
 However you ran `pip install`, running it again with `--update` will update to the latest release.
-For example, to update `requests` library:
+For example, to update the `requests` library:
 
 ```bash
 (project_venv) $ python -m pip install --update requests
--- a/tech/languages/python/pypi-installation.md
+++ b/tech/languages/python/pypi-installation.md
@@ -69,8 +69,6 @@ For example, to install a Python implementation of `cowsay` (rather than the ori
 $ python -m pip install --user cowsay
 ```
 
-If you're feeling lucky or impatient, you can use the shorter command `pip install --user cowsay`, without the `python -m`.
-Usually, this will install using the correct version of Python.
 
 ## Updating packages
 
--- a/tech/languages/python/pypi-installation.md
+++ b/tech/languages/python/pypi-installation.md
@@ -74,8 +74,9 @@ Usually, this will install using the correct version of Python.
 
 ## Updating packages
 
-Since software installed using `pip` are not part of Fedora, it is your responsibility to keep it updated as needed.
-However you ran `pip install`, running it again with `--update` will update to the latest release. For example:
+Since software installed using `pip` is not part of Fedora, it is your responsibility to keep it updated as needed.
+However you ran `pip install`, running it again with `--update` will update to the latest release.
+For example, to update `requests` library:
 
 ```bash
 (project_venv) $ python -m pip install --update requests


tech/languages/python/pypi-installation.md
```diff
@@ -56,16 +56,31 @@ If you no longer need it, deactivate it and delete it with `rm -rv project_venv`
...
+
+Then, use pip's `--user` option.
 For example, to install a Python implementation of `cowsay` (rather than the original Perl one available in Fedora), run:
 
 ```bash
 $ python -m pip install --user cowsay
 ```
 
-If you're feeling lucky or impatient, you can use the shorter command `pip install cowsay`, without the `python -m`.
+If you're feeling lucky or impatient, you can use the shorter command `pip install --user cowsay`, without the `python -m`.
```
pvalena, line 72: Could you explain the difference? Why not to use the shorter one as a default?

pvalena: Also, as this should be "shorter", isn't there a shorter arg for `--user`, like `-u`?

tech/languages/python/pypi-installation.md
```diff
@@ -56,16 +56,31 @@ If you no longer need it, deactivate it and delete it with `rm -rv project_venv`
...
 Usually, this will install using the correct version of Python.
 
+## Updating packages
+
+Since software installed using `pip` are not part of Fedora, it is your responsibility to keep it updated as needed.
+However you ran `pip install`, running it again with `--update` will update to the latest release. For example:
+
+```bash
+(project_venv) $ python -m pip install --update requests
```
pvalena, line 81: Please elaborate: what does `requests` mean in this context?

Also, I wanted to ask, in general- is there a simple way to "update all"?

pvalena (changes requested):
Thanks for the PR!

Please consider take a look at the comments, and address if possible. Feel free to merge it if you like either way.

tech/languages/python/pypi-installation.md
encukou: The difference is that there's no `python -m` at the start.
If this looks confusing, I'd remove the while sentence. It's mostly because you often see `$ pip install` in other guides; this hints that the wrong interpreter might get used when you take the shortcut.

There's no short form for `--user`.

tech/languages/python/pypi-installation.md
encukou: `requests` is the name of a package. This relates to the earlier command, `(project_venv) $ python -m pip install requests`.

There's no simple way to update all.

tech/languages/python/pypi-installation.md
pvalena: > The difference is that there's no `python -m` at the start.

Yes, off course, I meant the implications of that.

> If this looks confusing, I'd remove the while sentence. It's mostly because you often see `$ pip install` in other guides; this hints that the wrong interpreter might get used when you take the shortcut.

Maybe that's something worth mentioning in the guide.

tech/languages/python/pypi-installation.md
pvalena: > `requests` is the name of a package. This relates to the earlier command, `(project_venv) $ python -m pip install requests`.

Oh, it wasn't so apparent, as it was 40 lines above this. Maybe just a wording suggestion then (above).

tech/languages/python/pypi-installation.md
```diff
@@ -56,16 +56,31 @@ If you no longer need it, deactivate it and delete it with `rm -rv project_venv`
...
 ```
 
-If you're feeling lucky or impatient, you can use the shorter command `pip install cowsay`, without the `python -m`.
+If you're feeling lucky or impatient, you can use the shorter command `pip install --user cowsay`, without the `python -m`.
 Usually, this will install using the correct version of Python.
 
+## Updating packages
+
+Since software installed using `pip` are not part of Fedora, it is your responsibility to keep it updated as needed.
+However you ran `pip install`, running it again with `--update` will update to the latest release. For example:
```
pvalena, line 78:
```suggestion
However you ran `pip install`, running it again with `--update` will update to the latest release.
For example, to update `requests` library:
```

tech/languages/python/pypi-installation.md
```diff
@@ -56,16 +56,31 @@ If you no longer need it, deactivate it and delete it with `rm -rv project_venv`
...
 $ python -m pip install --user cowsay
 ```
 
-If you're feeling lucky or impatient, you can use the shorter command `pip install cowsay`, without the `python -m`.
+If you're feeling lucky or impatient, you can use the shorter command `pip install --user cowsay`, without the `python -m`.
 Usually, this will install using the correct version of Python.
 
+## Updating packages
+
+Since software installed using `pip` are not part of Fedora, it is your responsibility to keep it updated as needed.
```
pvalena, line 77: Little wording enhancement:

```suggestion
Since software installed using `pip` is not part of Fedora, it is your responsibility to keep it updated as needed.
```

tech/languages/python/pypi-installation.md
encukou: Is the following sentence, "Usually, this will install using the correct version of Python." not enough?

tech/languages/python/pypi-installation.md
encukou: Thanks for the suggestion!

tech/languages/python/pypi-installation.md
```diff
@@ -56,16 +56,32 @@ If you no longer need it, deactivate it and delete it with `rm -rv project_venv`
...
 
-If you're feeling lucky or impatient, you can use the shorter command `pip install cowsay`, without the `python -m`.
+If you're feeling lucky or impatient, you can use the shorter command `pip install --user cowsay`, without the `python -m`.
 Usually, this will install using the correct version of Python.
 
+## Updating packages
+
+Since software installed using `pip` is not part of Fedora, it is your responsibility to keep it updated as needed.
+However you ran `pip install`, running it again with `--update` will update to the latest release.
+For example, to update `requests` library:
```
encukou, line 79:
```suggestion
For example, to update the `requests` library:
```

tech/languages/python/pypi-installation.md
pvalena: Not sure. What it would do "unusually"? What happens if it uses incorrect version?

As someone who does not use python outside of virtualenv, I have no idea what are the implications :). Should I care?

pvalena: If best practise is always running it via `python -m`, I think we should stick with that.

hroncok: Mentioning also the option without it is good IMHO. Plus, it was already there (for a reason).

pvalena: > Mentioning also the option without it is good IMHO. Plus, it was already there (for a reason).

I'm not trying to persuade you to drop it, just trying to clear it up, as it's not quite straightforward what the actuall difference (**implications**) is. I really don't know. What is it good for (apart from being a bit shorter)?

Can you elaborate on the reason? As what I've actually gotten so far (check my question above), is that will "usually" do the correct "thing" which isn't much self-explanatory.

hroncok: The difference is:

 - `python -m pip install` allows you to pick the Python you nee explicitly (e.g. via `python3.8 -m pip`) and is preferred by us (as Python teachers)
 - `pip install` is easier to type and might be easier recognized by the readers who are already familiar with the command, but is less explicit and may sometimes lead to surprises

pvalena: Thanks for answering!

Well, If I may add one question: what happens if I'm in an virtualenv? Which one should I use? Or should I use the path to it's  `pip` binary?

For the latter use, could we clear up the sentence, like f.e.:

```suggestion
You can use the shorter command `pip install --user cowsay`, without the `python -m`, but first you should ensure you've set it to use correct python version.
```

Would that make sense?

hroncok: That really depends on what you mean by "I'm in an virtualenv".

That suggestion IMHO makes it worse, because it tells you to do something, without giving you a clue abut how to do it. (And we don't want to explain how to do this in this guide, because it could easily get out of our hands and grow to be a book.)

encukou: I removed the note.
Any way I can phrase it either raises more questions than it answers, or would be overwhelming the intended audience.

pvalena: > That really depends on what you mean by "I'm in an virtualenv".

To clarify:

using python3-virtualenv-20.4.4-1.fc35.noarch

``` bash
  $  virtualenv run 
  $  . run/bin/activate
```

Not sure if there's a difference for
```bash
  $ run/bin/pip install '.[all]' --upgrade
  # vs
  $ pip install '.[all]' --upgrade
  # vs
  $ python -m pip install '.[all]' --upgrade
```

And I'm probably not the only one who wonders :).

> 
> That suggestion IMHO makes it worse, because it tells you to do something, without giving you a clue abut how to do it. (And we don't want to explain how to do this in this guide, because it could easily get out of our hands and grow to be a book.)

Oh, I was hoping there would be an easy way to do so.

Maybe (in combination with the above) this would deserve a standalone page?

pvalena:
LGTM.

pvalena:
Thanks!

tech/languages/python/pypi-installation.md
encukou: They *should*  all be the same.

```
  $ run/bin/pip install '.[all]' --upgrade
```
This runs `pip` from the virtualenv explicitly. You don't need `. run/bin/activate` before it.
A slightly safer way to do it is `run/bin/python -m pip install '.[all]' --upgrade`, see below.

```
  $ pip install '.[all]' --upgrade
```
This works *if* nothing is wrong with your installation. If there's a mismatch between `python` and `pip` on your PATH, it'll install into a different location than your `python` command will search in.
This could happen for example if you uninstall `pip` from the virtualenv.

```
  $ python -m pip install '.[all]' --upgrade
```
This runs `pip` for the `python` in your PATH. You can't have the mismatch above if you only run the `python` command.

PR #389, opened 2021-07-16
Fixed bad link

qb42:
As of October 2016, https://fedoraproject.org/wiki/Mock says it is not maintained and directs users to https://github.com/rpm-software-management/mock/wiki instead.


jackorp:
Thanks!

PR #391, opened 2021-07-19
Update `postgresql-setup` command

samrobbins85:
When running the old command, it gives the output:

```
WARNING: using obsoleted argument syntax, try --help
WARNING: arguments transformed to: postgresql-setup --initdb --unit postgresql
```
This PR updates to the new argument syntax


pvalena:
Thanks, LGTM!

PR #396, opened 2021-09-24
Rename about.md to about.md5

Whatgraph:


jackorp:
Thanks for opening this PR however this change seems to be of no benefit. Closing.

PR #408, opened 2021-12-24
Revert "Update dead Python interpreters"

swansoffie001:
Reverts developer-portal/content#406


jackorp:
Hi, what is the motivation for this change?

The change was implemented in Fedora already: [Retire python3.5](https://fedoraproject.org/wiki/Changes/RetirePython3.5).

PR #418, opened 2022-02-18
Merged python django page to single page.

archijaiswal:
Please review the merged content and is this how it should be linked from languages section to web-app section?


jackorp:
Hi, Thanks for contributing!

I am very sorry to get back to you this late, there was a lot happening.

As to the PR:
I'd like to see also the "First project" and "Activating virtualenv" sections to be in the final merged file as that is good information and most other pages follow this path from installation to running the server, that's probably the biggest thing here.

As to how to link it, one alternative is to follow the PHP example: https://developer.fedoraproject.org/tech/languages/php/php-frameworks.html
However Leaving a page with just the link for now is fine as well.

Thinking into the future: it could have its own page on virtualenv alone that can be linked from these other guides, but let's keep it one thing at a time.

PR #422, opened 2022-03-11
Change long versions of no-doc for the short one

renich:
It's... shorter...


jackorp:
Hi, thanks for your contribution.

IMHO this boils down to preference. I'd rather have longer options that explain what they do than shorter options that I am, personally, bound to forget and will have to look them up.

There seems to be another problem with these and it's that these options seem outdated:

`gem help install` lists only `--no-document` and seems to not know about `--no-ri` or `--no-rdoc`

renich:
OK, we can change it to the long version then.

jackorp:
LGTM, thanks!

PR #423, opened 2022-03-11
gem installations for user by default

renich:
Added a method of asking rubygem to install it's gems to the user directory by default. Also added a suggestion to add the bin directory to the user's PATH.

### DIFF
--- a/tech/languages/ruby/ruby-installation.md
+++ b/tech/languages/ruby/ruby-installation.md
@@ -25,6 +25,20 @@ $ sudo dnf install rubygem-{tk{,-doc},rake,test-unit}
 
 Please note that we have already unbundled these libraries from Ruby itself, so they come in their own packages and need a specific dependency requirement in .gemspec or Gemfile as well as a specific `require()` call in your Ruby code.
 
+You might, also, want to install gems on your user's directory; as opposed to the system's, by default. If so, edit `$HOME/.gemrc` like this:
+
+```
+gem: --user-install
+```
+
+Remember to add: `HOME/.local/share/gem/ruby/bin` to your `$PATH`. One way of doing this is adding the following to `$HOME/.bash_profile`: 
+
+```
+# my path
+## ruby
+export PATH="$HOME/.local/share/gem/ruby/bin:$PATH"
+```
+
 ## JRuby
 
 Alternatively Fedora comes with JRuby packages that can be installed via:


jackorp:
Hi, thanks for your contribution.

I am not sure why is this needed. With Fedora's Ruby, gems are automatically installing "locally" for a user as opposed to system-wide.

I am also not sure about the bin path. When I install a gem (e.g. rspec) in a "fresh install" RubyGems generates executables that run the gem into the `~/bin/` directory and we do not need to add that into `$PATH`, default bashrc from the `bash` package has these lines included:
```bash
# User specific environment
if ! [[ "$PATH" =~ "$HOME/.local/bin:$HOME/bin:" ]]
then
    PATH="$HOME/.local/bin:$HOME/bin:$PATH"
fi
export PATH
```

IOW the directory that the gem executables end up in is already handled.
But I am not sure if it is handled by other shells.

jackorp:
Interesting thing: if you create/have a `$HOME/.gemrc` (empty one suffices), suddenly gems want to get installed system-wide. And the binpath is also different for `--user-install` under those circumstances...

renich:
> Interesting thing: if you create/have a `$HOME/.gemrc` (empty one suffices), suddenly gems want to get installed system-wide. And the binpath is also different for `--user-install` under those circumstances...

I was just going to check that. 

What do we do?

jackorp:
> What do we do?

To me, this seems like a bug, since we are not overriding the `install` options.

Reporting this in Bugzilla against Fedora Ruby would be a start.

I am one of maintainers of Fedora Ruby, I can report it and take a look later.

jackorp:
As to this PR's content, this does not seem like a bad idea.

I think that the more appropriate section would be IMO `gems-installation.md` as that part deals with gems.

And adjusting RubyGems behavior might be desirable whether you are running Ruby installed through DNF or rbenv or whatnot.

renich:
OK. I will move it to that then.

tech/languages/ruby/gems-installation.md
```diff
@@ -73,3 +73,19 @@ To look whether the required gem is available as an RPM package, look at
 [IsItFedoraRuby.com](http://isitfedoraruby.com/) site which tracks the Ruby
 integration within the Fedora project by listing available information about
 packaged gems.
+
+## Installing gems as a regular user
+
+You might want to install gems on your user's directory; as opposed to the system's, by default. If so, edit `$HOME/.gemrc` like this:
+
+```
+gem: --user-install
```
jackorp, line 82: Hmm, after reading up on this a bit,
there is another notation that does not override the globals:
```
install: --user-install
```
This preserves what the [operating_system.rb](https://src.fedoraproject.org/rpms/ruby/blob/rawhide/f/operating_system.rb#_103) sets.

with
```
$ echo "install: --no-documents" > ~/.gemrc
```
`gem env` then reports following:
```
     - "gem" => "--user-install --bindir /home/jprokop/bin" # <= default set by the mentioned `operating_system.rb`
     - "install" => "--no-document" # <= We set this previously in `.gemrc`
```

tech/languages/ruby/gems-installation.md
```diff
@@ -73,3 +73,19 @@ To look whether the required gem is available as an RPM package, look at
 [IsItFedoraRuby.com](http://isitfedoraruby.com/) site which tracks the Ruby
 integration within the Fedora project by listing available information about
 packaged gems.
+
+## Installing gems as a regular user
```
jackorp, line 77:
```suggestion
## Configuring RubyGems
```

tech/languages/ruby/gems-installation.md
jackorp: I'll leave it up to you whether or not to use it and leave the system's defaults intact (if left intact the `--bindir` that is set also stays and bins get generated to `$HOME/bin`)

jackorp (changes requested):
PTAL.

I am also concerned that it is not documenting Fedora's defaults or that it might be resetting them, which users might be using since we are in Fedora space. That might lead to unwanted confusion IMO.

tech/languages/ruby/gems-installation.md
renich: This is interesting. I didn't know this. Maybe it would be wise to mention all so that if the user creates `~/.gemrc` knows about said defaults. 

Thank you for all the corrections, btw.

renich: OK, it seems that my recommendation is useless if we use `install: --no-document`. I'll just fix that recommendation instead. 

Would it be worth leaving this as a notice, maybe? Just to let users know what happens when you use: `gem: --no-document` or such notation?

tech/languages/ruby/gems-installation.md
```diff
@@ -73,3 +73,19 @@ To look whether the required gem is available as an RPM package, look at
 [IsItFedoraRuby.com](http://isitfedoraruby.com/) site which tracks the Ruby
 integration within the Fedora project by listing available information about
 packaged gems.
+
+## Configuring RubyGems
+
+To install gems on your user's directory; as opposed to the system's, by default, edit `$HOME/.gemrc` like this:
```
pvalena, line 79: I don't think this is the case. In Fedora, we install to user directory by default, except for root (please don't use root for installing gems!).

https://gist.github.com/pvalena/af5241e975e4ad76be0f38f73f90d0a0

Or is there anything I've misunderstood about your intentions?

pvalena: You can also see defaults in the` gem install --help`:

```
 gem install --help | tail
    ==========================

    You can use `i` command instead of `install`.

      $ gem i GEMNAME

  Defaults:
    --both --version '>= 0' --no-force
    --install-dir /usr/share/gems --lock
    --document=ri

```

tech/languages/ruby/gems-installation.md
pvalena: +1 for preserving the system defaults and using `install:` instead.

tech/languages/ruby/gems-installation.md
renich: I changed my wording completely. It seems `gem: --no-document` will override system's defaults: https://src.fedoraproject.org/rpms/ruby/blob/rawhide/f/operating_system.rb#_103

... but, if we use: `install: --no-document`, it will not.

tech/languages/ruby/gems-installation.md
```diff
@@ -73,3 +73,19 @@ To look whether the required gem is available as an RPM package, look at
...
+```
+gem: --user-install
+```
+
+Remember to add: `$HOME/.local/share/gem/ruby/bin` to your `$PATH`. One way of doing this is adding the following to `$HOME/.bash_profile`: 
+
+```
+# my path
+## ruby
+export PATH="$HOME/.local/share/gem/ruby/bin:$PATH"
```
pvalena, line 103: AFAIK the symlinks get created in ${HOME}/bin for you. But I doubt it is default path for users.

tech/languages/ruby/gems-installation.md
jackorp: > OK, it seems that my recommendation is useless if we use `install: --no-document`. I'll just fix that recommendation instead.
> 
> Would it be worth leaving this as a notice, maybe? Just to let users know what happens when you use: `gem: --no-document` or such notation?

+1 for using `install: ...`. I like mentioning that `gem: ...` will override some defaults that Fedora has.

tech/languages/ruby/gems-installation.md
```diff
@@ -73,3 +73,32 @@ To look whether the required gem is available as an RPM package, look at
 [IsItFedoraRuby.com](http://isitfedoraruby.com/) site which tracks the Ruby
 integration within the Fedora project by listing available information about
 packaged gems.
+
+## Configuring RubyGems
+
+Fedora configures ruby to install gems on your user's directory; as opposed to the system's, by default. The gems get installed at: `$HOME/.local/share/gem`. 
+
+One can check the defaults [here](https://src.fedoraproject.org/rpms/ruby/blob/rawhide/f/operating_system.rb).
+
```
jackorp, line 82: Here is a good place to mention where is the config file or where to create/edit what we are talking about in this section ;)

tech/languages/ruby/gems-installation.md
```diff
@@ -73,3 +73,32 @@ To look whether the required gem is available as an RPM package, look at
...
+```
+Remember to add: `$HOME/.local/share/gem/ruby/bin` to your `$PATH` if you do since all settings aforementioned will be overriden. 
+
+One way of doing this is adding the following to `$HOME/.bash_profile`: 
+
+```
+# my path
+## ruby
+export PATH="$HOME/.local/share/gem/ruby/bin:$PATH"
+```
```
jackorp, line 104: Please also explain to the user what is this doing.

I.e. 'This will add the directory to the list of paths that shell is searching for binaries."

IOW something that is closer to this style: https://developer-portal.github.io/contributing/write-general-writing-hints.html#tell_the_user_whats_going_on

tech/languages/ruby/gems-installation.md
```diff
@@ -73,3 +73,32 @@ To look whether the required gem is available as an RPM package, look at
...
+
+One can check the defaults [here](https://src.fedoraproject.org/rpms/ruby/blob/rawhide/f/operating_system.rb).
+
+If one would like, say, to disable documentation generation at install time, one could do it in two ways:
+
+```
+install: --no-document
+```
+
+This method won't override the system's defaults.
```
jackorp, line 89:
```suggestion
This method will not override the system's defaults.
```

jackorp (changes requested):
A bit late mention, but better than never. JFTR we have writing tips for contributors on how to structure content: https://developer-portal.github.io/contributing/write-general-writing-hints.html

Please mention the `$HOME/.gemrc` file, it got a bit lost in the changes.

Try to simplify [line 96](https://github.com/developer-portal/content/pull/423/files#diff-b51576438039660a6f64d26aa1f8d9a11e5a2c4165e71994798b3223f4efa23aR96) I have a hard time understanding what it means.

Otherwise LGTM (if there would be any more outstanding issues we can iron them out later).

tech/languages/ruby/gems-installation.md
renich: I thought I did; in line 96.

renich:
:S I got lost as well...

tech/languages/ruby/gems-installation.md
voxik: I am not sure we can change this. The question is if the `help` command can reach somewhere here:

https://src.fedoraproject.org/rpms/ruby/blob/5bf57b1504871230600103083d77ff3502255e2e/f/operating_system.rb#_98

However, I am not mistaken, the way operating_system.rb acts similar to the options are specified on the command line and it would not make sense to change the help message in that case.

jackorp (changes requested):
LGTM. Last nit, somewhere in the changes, the mention of the configuration file (`~/.gemrc`) that we edit was lost. Please re-add that and it's good to merge.

renich:
Of course @jackorp. Sorry, I got lost with all the changes, heh. I'll fix it right away.

jackorp:
LGTM. Thanks!

PR #425, opened 2022-04-06
Align virtualenv advice for FastAPI with Django/Flask

### LINKED ISSUE #424: FastAPI content enhancement
From Ben Beasley on Fedora Devel list:

```
I noticed that the Flask and Django pages have language like:

    Fedora includes a python3-flask package that you can install and import. However, unless you are developing or packaging an application for Fedora, it is more useful to install Flask as a third-party package inside a virtual environment. This will keep your project separate from your system, giving you more freedom in choosing additional libraries and their versions, and easing collaboration with people who aren’t using Fedora yet.

which, I think is reasonable; for better or worse, this is how Python-based web development is conventionally done in practice.

It might be worthwhile to use similar language on the FastAPI page. Currently, the page recommends a virtualenv, but doesn’t mention that there is a python3-fastapi package in Fedora too.
```

musicinmybrain:
Specifically, mention that Fedora does contain a `python3-fastapi` package.

Fixes #424.

### DIFF
--- a/tech/languages/python/fastapi-installation.md
+++ b/tech/languages/python/fastapi-installation.md
@@ -10,9 +10,9 @@ order: 5
 
 ## Installation of FastApi in a virtual environment
 
-It is always useful to install FastApi as a third-party package inside a *virtual environemnt*.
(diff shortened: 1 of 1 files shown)


jackorp:
LGTM.

Thanks!

PR #429, opened 2022-04-29
updated the php server CLI commands

rotensepro:
Seems outdated, This works better !


jackorp:
Thanks for contribution.

We prefer long options to short ones for readability.

Issue #427, opened 2022-04-24
.Net IDEs : dotnet-ide.md


pvalena:
Link does not work:

https://developer.fedoraproject.org/deployment/Flatpak/about.html

Issue #407, opened 2021-12-24
: dotnet dotnet-ide.md


leonardon397:
When trying to install eclipse, I followed all the steps presented and even so I got the following message printed on my terminal below (it's in Portuguese):
```shell
[leonardo@fedora ~]$ sudo dnf install eclipse eclipse-mpc
Fedora 35 - x86_64                                   110 kB/s |  43 kB     00:00    
Última verificação de data de vencimento de metadados: 0:00:01 atrás em qui 23 dez 2021 19:32:31.
Sem correspondência para o argumento: eclipse
Sem correspondência para o argumento: eclipse-mpc
Erro: Impossível de encontrar uma correspondência: eclipse eclipse-mpc
```

Translating to english, it means
```shell
[leonardo@fedora ~]$ sudo dnf install eclipse eclipse-mpc
Fedora 35 - x86_64                                   110 kB/s | 43 kB     00:00
Last metadata expiration date check: 0:00:01 ago on Thu Dec 23, 2021 19:32:31.
No match for argument: eclipse
No match for argument: eclipse-mpc
Error: Unable to find a match: eclipse eclipse-mpc
```

Please what should I do to resolve this issue?

leonardon397:
Important info: I forgot to mention that, I am a Fedora 35 user, at the time I opened this issue.

jackorp:
Thanks for the report!

Hmm... it seems eclipse has been orphaned for some time[1], this will require a content update.

Alternatively, I was able to install eclipse and access the marketplace through the flatpak version.

```console
$ flatpak install org.eclipse.Java
```

[1] https://src.fedoraproject.org/rpms/eclipse/blob/rawhide/f/dead.package

leonardon397:
> Thanks for the report!
> 
> Hmm... it seems eclipse has been orphaned for some time[1], this will require a content update.
> 
> Alternatively, I was able to install eclipse and access the marketplace through the flatpak version.
> 
> ```
> $ flatpak install org.eclipse.Java
> ```
> 
> [1] https://src.fedoraproject.org/rpms/eclipse/blob/rawhide/f/dead.package

OK, I'm going to install from flatpak, until the problem is resolved!

PR #430, opened 2022-05-02
Replace dnf install with flatpak install for Eclipse dotNet IDE.

jackorp:
Eclipse was retired in Fedora 35, therefore it is not installable through
DNF.

Resolves: https://github.com/developer-portal/content/issues/407

Fix link in dotNet to flatpak pages.

Resolves: https://github.com/developer-portal/content/issues/427

Tested locally. I was able to install and run a smoke test of the aCute extension.


pvalena:
Can we remove the alternative downloads column and move it into the text instead?

pvalena:
Otherwise LGTM!

PR #439, opened 2022-06-09
Add content for Julia language

jakafle:
This PR adds instructions on getting started with Julia in Fedora. 

Thank you.


tech/languages/julia/julia-installation.md
```diff
@@ -0,0 +1,63 @@
...
+subsection: julia
+section: tech-languages
+version: 1.7.3
+order: 1
+description: High-level, high-performance dynamic language for technical computing.
+---
+
+# Julia installation and usage
+
+## Julia installation
```
jackorp, line 12: I think this is duplicating the content here. As there is no complicated usage I'd simply do a single h1 header

```suggestion
# Julia installation
```

Alternatively, do another h2 header after the installation steps.

jackorp (changes requested):
Mostly LGTM, please address the comment about the header.

jakafle:
I have updated the header with the suggestion you made. Thank you.

jackorp:
LGTM, Thanks!

PR #442, opened 2022-07-23
Correct the package irb comes from

matta:
Ruby's irb apparently used to be provided by the ruby package, but is now provided by rubygems-irb.


jackorp:
LGTM. Thanks!

PR #444, opened 2022-08-18
Draft: Update MongoDB installation and remove MongoDB GUI

jackorp:
It is not installable from Fedora since Fedora 30, but using the repo file seem to function OK.
Also update steps for enabling the systemd service.

MongoDB GUI repo is archived and I do not know of an alternative

### DIFF
--- a/tech/database/mongodb/about.md
+++ b/tech/database/mongodb/about.md
@@ -51,7 +51,3 @@ $ mongo
 MongoDB shell version: 4.0.0
 connecting to: test
 ```
-
-## MongoDB GUI (not tested)
-
-[MongoDB Admin](https://github.com/hatamiarash7/MongoDB_Admin/wiki/1.-Getting-Start) is a Cross Platform GUI.
--- a/tech/database/mongodb/about.md
+++ b/tech/database/mongodb/about.md
@@ -41,6 +41,8 @@ Now you can install with dnf
 $ sudo dnf install -y mongodb-org
 ```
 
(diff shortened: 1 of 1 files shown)


jackorp:
Hmm, the 4.4 version skips a few packages... There probably isn't a proper way to run it on Fedora other than a container maybe...

pvalena:
So there's no way to install this?

What about https://tecadmin.net/install-mongodb-on-fedora/ -- does it work?

jackorp:
4.4 works, with the exception of the `mongodb-org` `mongodb-org-database-tools-extra` `mongodb-org-tools` package, anything beyond that will not install due to this error present on both 6.0 and 5.0:
```
nothing provides /usr/libexec/platform-python needed by mongodb-org-database-tools-extra-5.0.9-1.el8.x86_64
```
And as noted in the other PRs, Fedora is not supported by MongoDB upstream. Other solutions might be: compile from source, use containers.

pvalena:
> 4.4 works, with the exception of the `mongodb-org` `mongodb-org-database-tools-extra` `mongodb-org-tools` package, anything beyond that will not install due to this error present on both 6.0 and 5.0:
> 
> ```
> nothing provides /usr/libexec/platform-python needed by mongodb-org-database-tools-extra-5.0.9-1.el8.x86_64
> ```
> 
> And as noted in the other PRs, Fedora is not supported by MongoDB upstream. Other solutions might be: compile from source, use containers.

I'm fine with whatever works (people use) ... so I'm fine with linking the blogpost (and mentioning it's a lower version, or recommending using CentOS (container), as that's supported by upstream. You might also mention the license is problematic.

jackorp:
https://github.com/developer-portal/content/pull/456 Let's continue here (I forgot I already made this PR...)

Issue #421, opened 2022-02-26
Groovy page needs update or removal.


jackorp:
Groovy is not in Fedora for some time now.

However, there is still a groovy page referencing `dnf install groovy` which is just not possible.

The page will either need new content that provides functional steps or be replaced with a page that explains the removal from Fedora or remove the page entirely.

@pvalena thoughts?

jackorp:
Moreover, the page uses incorrect yml headers.

https://github.com/developer-portal/content/blob/master/tech/languages/groovy/groovy-installation.md

PR #447, opened 2022-08-18
Deprecate groovy-installation.md.

jackorp:
Groovy is not available in Fedora since October 2019

Resolves: https://github.com/developer-portal/content/issues/421


pvalena:
Alternatives?

jackorp:
Installing from source or using some IDE might work... but what is actually usable for development with Groovy is hard to say...

pvalena:
Feel free to ask the experts on fedora-devel ML :).

jackorp:
Deprecation notice added instead

PR #449, opened 2022-08-18
Deprecate Cassandra page due to lack of maintanance.

jackorp:
Cassandra is not available on Fedora and
the page wasn't updated in a long time and contains quite an outdated content.


pvalena:
I'd replace this with header- something like:

```
## For reference purposes only - this page is outdated

Official web: https://cassandra.apache.org/_/index.html
```

PR #375, opened 2021-04-14
Mark Jython and Python 3.5 to be removed in Fedora 35

hroncok:
Related to https://github.com/developer-portal/content/issues/372

I've decided to keep the Jython info for the Fedora 34 release.

PR #450, opened 2022-08-25
Revert "Align virtualenv advice for FastAPI with Django/Flask (#425)"

musicinmybrain:
This reverts commit 63046c1e90f0f787eed5bace3ff0f1df440c3645.

The python-fastapi package has been retired due to lack of upstream support for Python 3.11.

### DIFF
--- a/tech/languages/python/fastapi-installation.md
+++ b/tech/languages/python/fastapi-installation.md
@@ -11,6 +11,9 @@ order: 5
 ## Installation of FastApi in a virtual environment
 
 Fedora includes a `python3-fastapi` package that you can install and import.
+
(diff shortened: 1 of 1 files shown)


jackorp:
Hi, thanks for the PR.

I don't think we need to fully revert these changes just yet. I'd consider some sort of deprecation note for now as it still will be installable on Fedora 36 and Fedora 35.

See https://github.com/developer-portal/content/pull/375 for an example.

musicinmybrain:
> Hi, thanks for the PR.
> 
> I don't think we need to fully revert these changes just yet. I'd consider some sort of deprecation note for now as it still will be installable on Fedora 36 and Fedora 35.
> 
> See #375 for an example.

How does the updated PR look to you?

jackorp:
Awesome, thanks!

PR #455, opened 2022-09-29
Update vagrant-libvirt page: qemu session

pvalena:
instructions, and also fixed upstream docs link.

### DIFF
--- a/tools/vagrant/vagrant-libvirt.md
+++ b/tools/vagrant/vagrant-libvirt.md
@@ -44,7 +44,7 @@ Read more on the `vagrant-libvirt` configuration in the [upstream documentation]
 
 ## Manage guest networks via Vagrant
 
-In Fedora `qemu://session` is used by default. This allows causes vagrant to run in userspace, not requiring password and being much safer. Downside is that not all Vagrant features are accessible, as they require root.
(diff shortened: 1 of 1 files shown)


tools/vagrant/vagrant-libvirt.md
```diff
@@ -44,10 +44,20 @@ Read more on the `vagrant-libvirt` configuration in the [upstream documentation]
 
 ## Manage guest networks via Vagrant
 
-In Fedora `qemu://session` is used by default. This allows causes vagrant to run in userspace, not requiring password and being much safer. Downside is that not all Vagrant features are accessible, as they require root. To revert for your selected boxes to `qemu://system`:
+In Fedora `qemu://session` is used by default. This allows causes vagrant to run in userspace, not requiring password and being much safer. Downside is that not all Vagrant features are accessible, as they require root.
```
pvalena, line 47:
```suggestion
In Fedora `qemu://session` is used by default. This allows causes vagrant to run in userspace, not requiring password and being much safer. Downside is that if some Vagrant functionality requires root permissions, it might not be available (some cases are resolved via "qemu bridge").
```

pvalena:
Please squash on merge.

PR #458, opened 2022-10-09
MongoDB is no longer free and open source.

kkofler:


jackorp:
Right, I overlooked that statement, thanks!

PR #459, opened 2022-10-11
arduino about page: style change

pvalena:
I'd rather not use the comment style, looks very strange, as if quoting something. Simple is better IMO.


jackorp:
The comment style looks a bit too aggressive

PR #467, opened 2022-12-27
Update vagrant-libvirt.md

Hofer-Julian:


tools/vagrant/vagrant-libvirt.md
```diff
@@ -51,9 +51,9 @@ To revert for your box to `qemu://system`, you can put something like the follow
 ```ruby
 Vagrant.configure("2") do |config|
   config.vm.provider :libvirt do |libvirt|
-    # Use QEMU session instead of system connection
-    libvirt.qemu_use_session = true
-
+    # Use QEMU system instead of session connection
+    libvirt.qemu_use_session = false
+end
```
jackorp, line 56: This was left without an end keyword on purpose. If we add the `end` it should be syntactically valid at least from Ruby POV.
```suggestion
    libvirt.qemu_use_session = false
  end
end
```

jackorp (changes requested):
Right, makes sense, if the session is the default then it probably makes sense to provide the actual instruction to use the system connection.

But let's make the code correct if it is not going to be open-ended.

tools/vagrant/vagrant-libvirt.md
Hofer-Julian: I trust you on this :)

jackorp:
LGTM. Thanks!

PR #469, opened 2023-01-22
Removed unsupported versions and added current .NET Core versions

CharDSonDev:
Versions 3.1 and 2.1 are no longer supported by Microsoft, they were replaced by the versions that are still supported.

More information in https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core.

The texts about not mixing Microsoft versions with Fedora versions were removed, since Microsoft stopped providing .NET Core versions for Fedora Linux.

### DIFF
--- a/tech/languages/dotnet/about.md
+++ b/tech/languages/dotnet/about.md
@@ -20,13 +20,6 @@ permalink: /tech/languages/csharp/dotnet-installation.html
 $ sudo dnf install dotnet
 ```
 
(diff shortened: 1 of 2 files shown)


jackorp:
A new, more broad PR on this topic cropped up: #471 , I will close this PR in favor of the more recent one if you agree.

CharDSonDev:
> 

I agree

PR #470, opened 2023-01-23
Added explanation for Fedora GOROOT

jboero:
I found this part problematic and not matching other distributions.

### DIFF
--- a/tech/languages/go/go-programs.md
+++ b/tech/languages/go/go-programs.md
@@ -20,7 +20,7 @@ If you get errors about missing packages, you may need to set a $GOROOT environm
 
 ```console
 $ sudo dnf install golang-src
-$ export GOROOT /usr/lib/golang
+$ export GOROOT=/usr/lib/golang
 ```
 
 Now, edit the file `hello.go` with your favorite editor and type the following:
--- a/tech/languages/go/go-programs.md
+++ b/tech/languages/go/go-programs.md
@@ -16,6 +16,13 @@ $ cd $HOME/go/src/hello
 $ touch hello.go
 ```
 
+If you get errors about missing packages, you may need to set a $GOROOT environment variable also. Fedora's `golang-src` package installs common packages into a GOROOT at `/usr/lib/golang` which should clear up most missing packages. Some distributions use `/usr/lib/golang` so this may be confusing for Fedora users reading docs from other distros. If you don't have root or sudo access, you may need to download your own [local GOROOT](https://go.dev/doc/manage-install) archive with basic packages into your home dir.
+
+```console
+$ sudo dnf install golang-src
+$ export GOROOT /usr/lib/golang
+```
+
 Now, edit the file `hello.go` with your favorite editor and type the following:
 
 ```go


jboero:
@cuviper or @voxik not sure who to assign this to but a simple docs addition for Go developers as I just wasted a few hours setting up a fresh GOROOT in latest F37.  Thanks!

tech/languages/go/go-programs.md
```diff
@@ -16,6 +16,13 @@ $ cd $HOME/go/src/hello
 $ touch hello.go
 ```
 
+If you get errors about missing packages, you may need to set a $GOROOT environment variable also. Fedora's `golang-src` package installs common packages into a GOROOT at `/usr/lib/golang` which should clear up most missing packages. Some distributions use `/usr/lib/golang` so this may be confusing for Fedora users reading docs from other distros. If you don't have root or sudo access, you may need to download your own [local GOROOT](https://go.dev/doc/manage-install) archive with basic packages into your home dir.
+
+```console
+$ sudo dnf install golang-src
+$ export GOROOT=/usr/lib/golang
+```
```
jackorp, line 24: As I said above if you follow the install that is a pre-requirement to this page, you would not need this.

IIRC golang is a statically compiled language. Therefore installing Golang libraries or Golang runtime of any kind is not needed for a built and distributed binary and is useless in an environment without a compiler. Am I correct in this assumption? If so, then what is your use-case of install just the golang-src?

jackorp (commented):
I'd appreciate reproduction steps to see the issue or even just logs.

Additionally, it'd be nice to put the specific error message in there so a reader can see what precisely is the message.

I am able to reproduce this if I go out of the GOPATH/src with the project. Then I run into some sort of GOROOT problem, not sure if it is the same.
```console
localhost $ podman run -it --rm fedora:37 bash
[root@cbca27f464c4 /]# dnf install golang -y && useradd -m test && su test
[test@cbca27f464c4 /]$ cd
[test@cbca27f464c4 ~]$ mkdir -p $HOME/go; echo 'export GOPATH=$HOME/go' >> $HOME/.bashrc; source $HOME/.bashrc
[test@cbca27f464c4 ~]$ mkdir -p $HOME/go/src/hello && cd $HOME/go/src/hello
[test@cbca27f464c4 hello]$ echo 'package main

import "fmt"

func main() {
	fmt.Println("Hello, Fedora!")
}
' > hello.go
[test@cbca27f464c4 hello]$ go mod init && go mod tidy && go build
go: creating new go.mod: module hello
go: to add module requirements and sums:
	go mod tidy
[test@cbca27f464c4 hello]$ ls
go.mod  hello  hello.go
[test@cbca27f464c4 hello]$ ./hello
Hello, Fedora!
```

Let's try adding custom dependencies and more custom code

```console
[test@cbca27f464c4 hello]$ go get github.com/hashicorp/go-retryablehttp
go: downloading github.com/hashicorp/go-retryablehttp v0.7.2
go: downloading github.com/hashicorp/go-cleanhttp v0.5.2
go: added github.com/hashicorp/go-cleanhttp v0.5.2
go: added github.com/hashicorp/go-retryablehttp v0.7.2
[test@cbca27f464c4 hello]$ cat hello.go
package main

import "fmt"
import "github.com/hashicorp/go-retryablehttp"

func main() {
	resp, err := retryablehttp.Get("/foo")
		if err != nil {
			panic(err)
		}
	fmt.Println(resp)
	fmt.Println("Hello, Fedora!")
}
[test@cbca27f464c4 hello]$ go build
[test@cbca27f464c4 hello]$ ./hello
2023/01/23 13:12:20 [DEBUG] GET /foo
2023/01/23 13:12:20 [ERR] GET /foo request failed: Get "/foo": unsupported protocol scheme ""
panic: GET /foo giving up after 1 attempt(s): Get "/foo": unsupported protocol scheme ""

goroutine 1 [running]:
main.main()
	/home/test/go/src/hello/hello.go:9 +0xc5
```

So, custom dependency was picked up as expected.

However, if I do not use the github.com/bla/bla notation and instead I use `import "retryablehttp"`, then I see this:
```console
[test@cbca27f464c4 hello]$ go build
hello.go:4:8: package retryablehttp is not in GOROOT (/usr/lib/golang/src/retryablehttp)
```

However, that does not seem to be the problem of the installation or our instructions, but problem on the side of the developer.

jackorp:
Sorry, I had to delete a few comments of mine, they were duplicated. GitHub saved already-made comments when I tabbed out but I did not notice it and thought they are gone...

jboero:
Thanks for the feedback. I see things seem to have changed for me since last time I set up a fresh Go env. I had an explicit GOROOT env variable set. Removing that the platform correctly defaults to /usr/lib/golang.

Yes `golang-src` should be installed as a dependency but it's not clear that that's the package that includes a Fedora-signed GOROOT in path `/usr/lib/golang`. All of GoLang's public documentation is Ubuntu-oriented or at least avoids packaging saying to just expand an archive into `/usr/local/go`. I see conflicting threads in old and new Go docs and forums that say whether a GOROOT or GOPATH must be set a certain way. The latest Fedora defaults seem fine so long as you have root access to install the standard packaging rather than download a goroot archive.

https://go.dev/doc/install

Up to you if you think this helps now. I see it's not really necessary.

tech/languages/go/go-programs.md
jboero: In this case external github imports weren't working because of failed dependencies but I had a custom GOROOT set which wasn't right. Fedora default seems fine if golang-src is installed otherwise non-root users can download an archive of go for a local GOROOT.

jackorp:
Glad to hear it works OK from the default Fedora package. If it'd been otherwise, I'd consider it a Fedora bug :).

BTW it seems that with the Fedora package, it is possible to have different versions available. You can follow this upstream guide: https://go.dev/doc/manage-install 

Since you no longer see these changes as necessary, I'll close this PR.

PR #481, opened 2023-06-18
Added info about mod_wsgi & dependencies

SomeAB:
I added further information about how to install mod_wsgi first using pip venv method, and related dependencies as of Python 3.11, pip 23.1, mod_wsgi 4.9.4

Original pypi page briefly mentions python3-devel as a dependency, but missing a mention of the package name, when talking about Fedora related dependencies. 

This just adds more clarity, if someone wants to install mod_wsgi before installing django.

### DIFF
--- a/tech/languages/python/django-installation.md
+++ b/tech/languages/python/django-installation.md
@@ -29,6 +29,14 @@ To work in the virtual environment, you have to activate it.
 $ source project_venv/bin/activate
 ```
 
+## How to install mod_wsgi in a virtual environment before installing Django
+
+```bash
+(project_venv) $ pip install wheel
+(project_venv) $ pip install mod_wsgi
+```
+Without installalling 'wheel' first, you will get a warning about legacy install method. See https://github.com/pypa/pip/issues/8559 for more details. Though as of pip 23.1, pythong 3.11, it should still install correctly without wheel. Additional dependencies 'may' include python3-devel, httpd, httpd-devel, setuptools, redhat-rpm-config. 
+
 In an active the virtual environment (with the name `(project_venv)` included in your command line prompt), you can install Django [from PyPI](https://developer.fedoraproject.org/tech/languages/python/pypi-install.html).
 
 ```bash


tech/languages/python/django-installation.md
```diff
@@ -29,6 +29,14 @@ To work in the virtual environment, you have to activate it.
 $ source project_venv/bin/activate
 ```
 
+## How to install mod_wsgi in a virtual environment before installing Django
+
+```bash
+(project_venv) $ pip install wheel
+(project_venv) $ pip install mod_wsgi
+```
+Without installalling 'wheel' first, you will get a warning about legacy install method. See https://github.com/pypa/pip/issues/8559 for more details. Though as of pip 23.1, pythong 3.11, it should still install correctly without wheel. Additional dependencies 'may' include python3-devel, httpd, httpd-devel, setuptools, redhat-rpm-config.
```
jackorp, line 38: Typo
```suggestion
Without installalling 'wheel' first, you will get a warning about legacy install method. See https://github.com/pypa/pip/issues/8559 for more details. Though as of pip 23.1, python 3.11, it should still install correctly without wheel. Additional dependencies 'may' include python3-devel, httpd, httpd-devel, setuptools, redhat-rpm-config. 
```

jackorp, line 38: I don't really like the "may". Does it or does it not? If it does, it should be explicitly mentioned with a `dnf install` command, if it does not, it doesn't need to be mentioned.

jackorp (changes requested):
Mostly good, though I'd prefer if the dependencies were known for sure.

Thanks.

SomeAB:
@jackorp okay, I will again check both 'source' build method & pip install method for dependencies. I agree. It should be clearer. Will post back in 2-3 days.

SomeAB:
I just checked this properly (sorry for late reply)

On the https://developer.fedoraproject.org/tech/languages/python/django-installation.html itself .. the above stated information might not be suitable

Modwsgi is a replacement for the outdated CGI (CGI is included in Apache by default). It can help run other frameworks as well like Flask. Modwsgi is also the recommended WSGI interface for Django (according to official Django Documentation)

I think it would be better to make a page, linked on the left side menu to add information about Modwsgi (i.e., it deserves a page of its own). 

However, there seems to be no way to raise a pull request for making a separate page (& might need manual intervention of the wiki moderator/admin)

If you feel like it, you can close this pull request

jackorp:
> However, there seems to be no way to raise a pull request for making a separate page (& might need manual intervention of the wiki moderator/admin). See below image:

Hmm... the closest I can find is https://developer-portal.github.io/doc/file-structure.html and https://developer-portal.github.io/doc/structure.html
you can basically copy-paste a page in the correct directory (there will probably be the other pages as well), edit the YAML on the top and write the correct contents.
Of course I can check & correct any wrong values, etc...

But I am not sure we have complete documentation for that process.

> If you feel like it, you can close this pull request

If you judge the content of this PR unsuitable, as described, in the current state, I will. I would of course welcome the contribution on a separate page. If you would like help with continuing on that separate page, please create an issue for this repo where we can collaborate on this.

Thanks.

PR #492, opened 2023-11-15
Update rails.md to use packaged rbenv

ekohl:
This updates the document to use the packaged rbenv instead of installing it from git. It also updates the version numbers to be more recent.

epel-release installation is dropped since that only applies to Enterprise Linux. Even there EL8+ has nodejs packaged.

### DIFF
--- a/start/sw/web-app/rails.md
+++ b/start/sw/web-app/rails.md
@@ -6,33 +6,24 @@ order: 6
 
 # Installing Ruby and Rails with rbenv
 
-The first step is to install dependencies for Ruby.
+The first step is to install dependencies for Ruby and rbenv.
 
 ```
-sudo dnf install git-core zlib zlib-devel gcc-c++ patch readline readline-devel libyaml-devel libffi-devel openssl-devel make bzip2 autoconf automake libtool bison curl sqlite-devel
+sudo dnf install git-core zlib zlib-devel gcc-c++ patch readline-devel libyaml-devel libffi-devel openssl-devel make bzip2 autoconf automake libtool bison libcurl-devel sqlite-devel rbenv rbenv-build-rbenv
 ```
 
-Install rbenv
+Then configure your shell to enable rbenv:
 
 ```
-cd
-git clone https://github.com/rbenv/rbenv.git ~/.rbenv
-echo 'export PATH="$HOME/.rbenv/bin:$PATH"' >> ~/.bashrc
 echo 'eval "$(rbenv init -)"' >> ~/.bashrc
 source ~/.bashrc
-exec $SHELL
-
-git clone https://github.com/rbenv/ruby-build.git ~/.rbenv/plugins/ruby-build
-echo 'export PATH="$HOME/.rbenv/plugins/ruby-build/bin:$PATH"' >> ~/.bashrc
-source ~/.bashrc
-exec $SHELL
 ```
 
-Install Ruby
+Install a Ruby version, such as 3.1.2. See `rbenv install --list` for available versions.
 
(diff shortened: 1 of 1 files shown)


start/sw/web-app/rails.md
```diff
@@ -6,33 +6,24 @@ order: 6
 
 # Installing Ruby and Rails with rbenv
 
-The first step is to install dependencies for Ruby.
+The first step is to install dependencies for Ruby and rbenv.
 
 ```
-sudo dnf install git-core zlib zlib-devel gcc-c++ patch readline readline-devel libyaml-devel libffi-devel openssl-devel make bzip2 autoconf automake libtool bison curl sqlite-devel
+sudo dnf install git-core zlib zlib-devel gcc-c++ patch readline-devel libyaml-devel libffi-devel openssl-devel make bzip2 autoconf automake libtool bison libcurl-devel sqlite-devel rbenv rbenv-build-rbenv
```
jackorp, line 12: I think there are a few dependencies missing, the updated instructions here https://developer.fedoraproject.org/tech/languages/ruby/ruby-installation.html list more things. I am quite sure that without, at least, the Perl dependencies, Using some Ruby Versions will fail without them.

jackorp (changes requested):
Mostly LGTM, some smaller notes. PTAL. Thanks.

start/sw/web-app/rails.md
```diff
@@ -42,12 +33,6 @@ Use this command if you do not want rubygems to install the documentation for ea
...
 
-Install bundler
-
-```
-gem install bundler
-```
-
 Whenever you install a new version of Ruby or a gem, you should run the rehash sub-command. This will make rails executables known to rbenv, which will allow us to run those executables:
 
 ```
```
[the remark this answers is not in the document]
ekohl: I had the same thought. Not sure where we should land.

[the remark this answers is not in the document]
jackorp: Eh. I think it was removed from the upstream instructions for Ruby language as it got updated, but I don't think I was testing out actual gem library executables and I don't feel like doing the whole rbenv setup from beginning again. If you feel like testing whether it actually is required, feel free to.

start/sw/web-app/rails.md
ekohl: Hmm, it's been a while since I ran it on my desktop and I recall having some perl issues but then I just ran it in a container and it worked without.

Though now that I look at the bigger picture, why don't we simply instruct users for Rails to follow https://developer.fedoraproject.org/tech/languages/ruby/ruby-installation.html first? Or use the system installed Ruby by default?

jackorp: The perl issues are almost Ruby version specific. (yes, Ruby, not rbenv...)

Instructing users to follow that one is also a possibility, I wanted to suggest de-duplicating this way, but then I read you actually use the system rbenv, which we don't describe on that page. It can be of course brought over and just linked here for users to first follow those instructions.

And I would recommend against using system Ruby for Rails. In my entirely personal opinion, working with Rails without rbenv or other way to keep some semblance of a stable environment for more than a year gets very annoying.

start/sw/web-app/rails.md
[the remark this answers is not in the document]
jackorp: Rehash is currently needed for bundler. See #494

start/sw/web-app/rails.md
jackorp: Let's leave this be for now and think about it at some later PR for rbenvs.

Issue #494, opened 2023-11-28
: web-app rails.md


dirschn:
I noticed today that the Ruby on Rails guide mentions
> Whenever you install a new version of Ruby or a gem, you should run the rehash sub-command. This will make rails executables known to rbenv, which will allow us to run those executables:

I was a little confused because I haven't manually run that  command ever and haven't had any issues, so I took a look at the [rbenv GitHub page](https://github.com/rbenv/rbenv) and in the [`rbenv rehash`](https://github.com/rbenv/rbenv#rbenv-rehash) section it mentions that `rbenv rehash` is run automatically when installing gems. It seems to me that there's no need to run it manually after installing new versions or gems?

Apologies if I have the format of creating a new issue wrong, I couldn't find any format requirements!

jackorp:
Thanks for taking time sharing this!

Yeah we were discussing this recently on some rbenv related pull request whether that's needed, we left it there due to not being sure. Thanks for finding the reference that it isn't needed, We can fix that with confidence now.

Alternatively, you can edit the page yourself here: https://github.com/developer-portal/content/blob/master/start/sw/web-app/rails.md , we have some related info for contributing here: https://developer-portal.github.io/contributing/

benmelz:
Coworker of @dirschn here, with a small gotcha. I'm observing with `rbenv 1.2.0-80-gd6bb1c6` that while `gem install` appears to trigger automatic rehashing, `bundle` does not.

We include the following in our setup script and haven't added cli tools in a while, hence @dirschn's confusion
```ruby
  system('bundle check') || system!('bundle install')
  system('if command -v rbenv &> /dev/null; then rbenv rehash; fi')
```

Since the majority of rails devs will be installing tools via `bundle`, that might warrant leaving this section in for now at least (perhaps open an issue with the rbenv team to expand functionality to `bundle`? and then this could be removed).

jackorp:
Huh, that's interesting behavior. So to get my mental notes on this straight:

The following triggers a rehash:
```
gem install foo-bar-with-executable
```

While the following command in a project with a Gemfile containing entry for gem `foo-bar-with-executable`:
```
bundle install
```
will not then execute rehash and therefore `foo-bar-with-executable` will not be available correctly?

I am a little bit confused as I have not yet had the need for bundler + rbenv combo,
is it regularly expected that bundle install would make `foo-bar-with-executable` available on regular path with rbenv after a rehash?

A rewording or similar might be warranted though if it works for all but bundler environments.

[...]

PR #498, opened 2023-12-20
Revitalize web-app section about Rails.

jackorp:
Fixes: https://github.com/developer-portal/content/issues/494

### DIFF
--- a/start/sw/web-app/rails.md
+++ b/start/sw/web-app/rails.md
@@ -42,30 +42,17 @@ $ echo "gem: --no-ri --no-rdoc" > ~/.gemrc
 
 ## Rbenv rehash
 
-Running rbenv rehash is not needed for simple `gem install`. rbenv triggers a rehash automatically.
-It may be desired to rehash after installing gems with executables via bundler.
-It will make the Ruby gem executables known to rbenv, which allows running those executables.
-
-You can rehash by executing the following:
-
+`rbenv rehash` is not needed when installing gems via `gem install` as rbenv will automatically
+trigger the rehash. You may need to trigger a rehash manually after installing gems using bundler.
+You can do so by running
 ```console
 $ rbenv rehash
 ```
 
-Since [rbenv commit 325abac](https://github.com/rbenv/rbenv/commit/325abac17de79a230152bb7038126a0641c6aa64)
-the `rbenv rehash` command is not required when using bundler.
-A rehash will trigger automatically similarly to running `gem install`.
-
-If you have cloned rbenv from upstream ensure you have the latest rbenv that contains the fix.
-You can update rbenv pulled from upstream using git.
-
-Assuming you have cloned your rbenv to `~/.rbenv`, enter rbenv's directory and pull latest updates
-from the upstream git repository:
-
-```console
-$ cd ~/.rbenv
-$ git pull
-```
+Since [rbenv commit 325abac](https://github.com/rbenv/rbenv/commit/325abac17de79a230152bb7038126a0641c6aa64),
+there is no need to run `rbenv rehash` when installing gems via bundler or `gem install`.
+Rbenv will automatically trigger the rehash using either of those methods. To ensure you have this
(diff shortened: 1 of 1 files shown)


start/sw/web-app/rails.md
```diff
@@ -8,70 +8,95 @@ order: 6
...
 ```
 Use this command if you do not want rubygems to install the documentation for each package locally.
 
-```
+```console
 echo "gem: --no-ri --no-rdoc" > ~/.gemrc
 ```
 
-Whenever you install a new version of Ruby or a gem, you should run the rehash sub-command. This will make rails executables known to rbenv, which will allow us to run those executables:
+## Rbenv rehash
```
dirschn, line 43: Github won't let me leave a suggestion on all of it because some of it has been deleted, so here's my suggestion to make this section flow a little better:

---
`rbenv rehash` is not needed when installing gems via `gem install` as rbenv will automatically trigger the rehash. You may need to trigger a rehash manually after installing gems using bundler. You can do so by running 

```console
$ rbenv rehash
```

Since [rbenv commit 325abac](https://github.com/rbenv/rbenv/commit/325abac17de79a230152bb7038126a0641c6aa64), there is no need to run `rbenv rehash` when installing gems via bundler or `gem install`. Rbenv will automatically trigger the rehash using either of those methods. To ensure you have this version installed, follow the [basic git checkout installation instructions](https://github.com/rbenv/rbenv?tab=readme-ov-file#basic-git-checkout)

---

I don't think we need to include instructions on how to update/install rbenv, the link to their instructions is probably better.

dirschn (changes requested):
I personally am not a fan of including `$` at the beginning of the console lines, but I recognize and concede that that is a pretty standard practice these days.

start/sw/web-app/rails.md
```diff
@@ -8,70 +8,95 @@ order: 6
 
 The first step is to install dependencies for Ruby and rbenv.
 
+```console
+$ sudo dnf install git-core zlib zlib-devel gcc-c++ patch readline-devel libyaml-devel libffi-devel openssl-devel make bzip2 autoconf automake libtool bison libcurl-devel sqlite-devel perl-FindBin perl-lib perl-File-Compare
```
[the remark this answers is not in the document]
jackorp: Last time I went over installing upstream supported Rubies, these were needed for Ruby 3.0: https://github.com/developer-portal/content/pull/489#discussion_r1348807772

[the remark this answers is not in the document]
jackorp: Technically they would be pulled in by git I think? But sice we are using git-core, these Perl deps are skipped.

start/sw/web-app/rails.md
jackorp: I hope I copy pasted right :)

jackorp:
Thanks a lot for taking a look. Suggestions applied, please re-review.

start/sw/web-app/rails.md
```diff
@@ -8,70 +8,82 @@ order: 6
...
-rbenv rehash 
-``` 
+Since [rbenv commit 325abac](https://github.com/rbenv/rbenv/commit/325abac17de79a230152bb7038126a0641c6aa64),
+there is no need to run `rbenv rehash` when installing gems via bundler or `gem install`.
+Rbenv will automatically trigger the rehash using either of those methods. To ensure you have this
+version installed, follow the [basic git checkout installation instructions](https://github.com/rbenv/rbenv?tab=readme-ov-file#basic-git-checkout).
 
 # Installing Rails
 
 Rails depends on a Javascript runtime, install nodejs.
```
dirschn, line 59: A thought I had, since `rbenv` is suggested for installing Ruby it could be nice at least mention `nodenv`? I think for Fedora the best option is [installing using git](https://github.com/nodenv/nodenv?tab=readme-ov-file#basic-github-checkout)

jackorp: Not a bad idea, created a new issue for it #502 , but let's move current PR through if possible.

dirschn (approved):
Looks good to me

jackorp:
Thanks for review!

PR #539, opened 2024-05-04
[Rust] Expand section about cargo + crate dependencies

decathorpe:
Rust crates as packaged by Fedora are not really a suitable development target - this is one of the reasons why using them as a replacement for crates.io is not a supported use case.

While it was intentional that there was no documentation for how to use the Fedora packages for Rust crates for local development (it is not a supported use case, after all), this has caused confusion on multiple occasions. To remove this ambiguity, this commit explicitly documents this use case as unsupported in the developer documentation.

### DIFF
--- a/tech/languages/rust/rust-cargo.md
+++ b/tech/languages/rust/rust-cargo.md
@@ -43,10 +43,22 @@ $ cargo run
 Hello, world!
 ```
 
-# crates.io repository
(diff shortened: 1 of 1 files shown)


jackorp:
Sorry for the delay, just got back from time-off, good stuff, thanks. LGTM.

decathorpe:
No problem. Thanks for merging!

PR #542, opened 2024-06-22
Mention copr-cli video tutorial

FrostyX:
The direct link to the playlist is here:
https://www.youtube.com/playlist?list=PLEOLABfRmCVjhTCrxiHNE9KwDepnbhIXa

Or we can point to the documentation instead:
https://docs.pagure.org/copr.copr/video_tutorial.html#video-tutorial


jackorp:
Nice, thanks. The playlist will should be enough, I don't want to indirect it one more level.

PR #547, opened 2024-11-16
Updated a command in docker-installation.md

hatwearingdumb:
Updated a command. 

The previous one returns the following error:
> Unknown argument "--add-repo" for command "config-manager". Add "--help" for more information about the arguments.

Tested on Fedora 41.

### DIFF
--- a/tools/docker/docker-installation.md
+++ b/tools/docker/docker-installation.md
@@ -17,11 +17,15 @@ $ sudo dnf install dnf-plugins-core
 ```
 
 To add the `docker-ce` repository
+
 * on Fedora 40:
+
 ```console
 $ sudo dnf config-manager --add-repo https://download.docker.com/linux/fedora/docker-ce.repo
 ```
+
 * on Fedora 41 and newer:
+
 ```console
 $ sudo dnf config-manager addrepo --from-repofile="https://download.docker.com/linux/fedora/docker-ce.repo"
 ```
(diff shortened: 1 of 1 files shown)


tools/docker/docker-installation.md
```diff
@@ -19,7 +19,7 @@ $ sudo dnf install dnf-plugins-core
 To add the `docker-ce` repository
 
 ```console
```
jackorp, line 21: That is the correct form for dnf5, however, Fedora 40 that's still supported where dnf4 is still default (at least in registry.fedoraproject.org/fedora:40 container), the original form is the correct one.

````suggestion
* on Fedora 40:
```
$ sudo dnf config-manager --add-repo https://download.docker.com/linux/fedora/docker-ce.repo
```
* on Fedora 41 and newer:
```console
````

Not sure whether using a form like `* on dnf4 (Fedora 40):` is better or too much

jackorp:
PTAL at the comment, thanks for the contribution!

tools/docker/docker-installation.md
hatwearingdumb: I figured that could be the origin of the change, but thanks for confirming! Definitely a good idea to list both.
> Not sure whether using a form like * on dnf4 (Fedora 40): is better or too much

I think stating the OS' version is enough.

jackorp:
LGTM, thanks!

PR #548, opened 2024-11-16
Update docker installation options

buckaroogeek:
- Amends developer documentation to include a link to the Docker and Docker Composes installation guide in Fedora Quick Docs.
- Amends Compose description with an overview of important changes in Compose V2.

### DIFF
--- a/tools/docker/compose.md
+++ b/tools/docker/compose.md
@@ -6,7 +6,7 @@ order: 5
 
 # Docker Compose
 
-[Docker Compose](https://docs.docker.com/compose/) is a tool to orchestrate Docker one or more containers using a simple YAML file which describes your whole setup.
+[Docker Compose](https://docs.docker.com/compose/) is a tool to orchestrate one or more containers using a simple YAML file which describes your whole setup.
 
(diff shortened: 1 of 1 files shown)


tools/docker/compose.md
```diff
@@ -6,8 +6,9 @@ order: 5
 
 # Docker Compose
 
-[Docker Compose](https://docs.docker.com/compose/) is a tool to orchestrate Docker containers using a simple YAML file which describes your whole setup.
+[Docker Compose](https://docs.docker.com/compose/) is a tool to orchestrate Docker one or more containers using a simple YAML file which describes your whole setup.
```
jackorp, line 9: "orchestrate Docker one or more" does not seem to be in correct order

Would the following work here?
```suggestion
[Docker Compose](https://docs.docker.com/compose/) is a tool to orchestrate one or more Docker containers using a simple YAML file which describes your whole setup.
```

Not sure whether docker-compose works with podman. If it's sensible to say that it works in general with containers, then we could throw out the "Docker" word.

jackorp (changes requested):
PTAL on the comment, otherwise content looks good, thanks!

tools/docker/compose.md
buckaroogeek: Thanks.  docker-compose will work with podman if podman-docker is installed.  I recommend 

`..to orchestrate one or more containers using a simple YAML file ...`

thank you

jackorp:
LGTM, Thanks!

jackorp (approved):

PR #550, opened 2024-12-19
Update .NET docs, replacing .NET 6 with .NET 9

### LINKED ISSUE #549: .NET docs reference EOL v6, don't include STS v9
At present, `tech/languages/dotnet/dotnetcore.md` and `tech/languages/dotnet/about.md` references .NET 6 and .NET 8.

Fedora 40 provides 6, 8, and 9 while Fedora 41 provides 8 and 9. With 6 being EOL and the Fedora documentation mentioning versions supported by all supported releases, 6 should be removed and 9 should be added.

rnelson:
Removes .NET 6 and adds .NET 9. Fixes #549.

### DIFF
--- a/tech/languages/dotnet/about.md
+++ b/tech/languages/dotnet/about.md
@@ -3,7 +3,7 @@ title: .NET Platforms
 subsection: dotnet
 section: tech-languages
 order: 1
-version: 7.8.4
+version: 7.8.5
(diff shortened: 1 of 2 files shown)


jackorp (changes requested):
LGTM, only see comment I made on 1 line after that we can merge it.

jackorp (approved):
Thanks!

PR #551, opened 2025-01-07
 Update redis to valkey as it's the new project in fedora 41

nihed:


nihed:
Got it, thank you for these information

PR #552, opened 2025-01-23
Update Python versions and remove old notice

xanderlent:
I happened to glance at the Python documentation and noticed it was outdated.

I've tried to fix those things which I have noticed, specifically:

- The list of available Python versions
- The old deprecation notice for fastapi; it is still packaged in Fedora


jackorp:
LGTM, thanks!

PR #567, opened 2025-08-27
Update the commands to work with new esptool, keep compatibility

befeleme:
esptool 5.0.0 has changed the command and option names (uses "-" instead of "_").
Adjust the commands in the article, and also keep the old options for compatibility.


jackorp:
LGTM. Thanks!

PR #568, opened 2025-08-27
Remove pipenv page

befeleme:
pipenv is not currently available in any Fedora version.

@hroncok, WDYT? Should we get rid of it altogether?


hroncok:
> Should we get rid of it altogether?

Yes please.

hroncok (approved):

PR #570, opened 2025-08-27
Update nodejs.md

tjuhaszrh:
Remove mentions of modules from nodejs.md and replace it with current method for installing alternative streams

### DIFF
--- a/tech/languages/nodejs/nodejs.md
+++ b/tech/languages/nodejs/nodejs.md
@@ -21,21 +21,13 @@ This will install V8 Javascript Engine, Node.js runtime and npm package manager 
 
 ### Alternative versions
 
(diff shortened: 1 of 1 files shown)


tech/languages/nodejs/nodejs.md
```diff
@@ -63,21 +55,25 @@ Installing Node.js modules is covered in [Node.js modules](/tech/languages/nodej
 ## Installing Global Modules
 
 Create a directory for global installations inside your home directory:
+
 ```
 mkdir ~/.npm-global
 ```
 
 Set the new directory path for npm:
+
```
jackorp, line 64: Would prefer these in a separate commit next time.

